PR #1, opened 2017-09-01
Use heading syntax in Markdown

scpeters:
Adding a space between #* and text causes text to render as headings. This is subjective, so it's up to you of course.

Preview: https://github.com/scpeters/astrobee/blob/0b5e529360d92c1f138d531af20713e93ff3bcc0/simulation/readme.md


marinagmoreira:
This is corrected in newer versions.
Thanks for reporting.

PR #3, opened 2018-06-28
Fix link to guest_science.md

Jkillelea:
Link was pointing to `...astrobee_android/blob/HEAD/guest_science.md` instead of `...astrobee_android/blob/master/guest_science.md`


marinagmoreira:
Hi Jacob!
This issue is corrected in newer versions, thanks for reporting.

PR #11, opened 2019-06-27
fixed (updated) ROS debian package repo GPG key

astrobotic-maxgarber:
c.f. http://answers.ros.org/question/325039/apt-update-fails-cannot-install-pkgs-key-not-working/#js-post-body-325040


marinagmoreira:
Hi!
The debian GPG key was updated in the last release.
Thanks for reporting!

PR #20, opened 2020-03-27
add memory fading of inflated octomap

Kevinkald:
Added memory fading of the inflated octomap (tree_inflated_). Previosly this tree_inflated_ would grow on the order of ten times as large as the regular octomap (tree_). This has been tested in simulations in iss.world.

Also added pruning of the tree_inflated_ tree.

Fixed two typos.


marinagmoreira:
Hi Kevin.
All your pull requests were merged in release v0.11.0.
Thank you so much for all the dedicated time, we really appreciate it!
We'll hopefully continue the collaboration in the future :)

PR #21, opened 2020-03-27
add check if Astrobee is dynamically stopped in truth mode

Kevinkald:
Added check for determining if Astrobee is dynamically stopped in truth mode. Previously the check was only done in the ekf, i.e when not using truth mode. This pull request should only have changed the TwistCallback function in gnc/ctl/src/ctl.cc, compared to the pull request earlier for the controller.

Changes has been tested in simulation using the iss.world, by initiating a stop command for the controller both in truth mode and not in truth mode.


marinagmoreira:
Hi Kevin.
All your pull requests were merged in release v0.11.0.
Thank you so much for all the dedicated time, we really appreciate it!
We'll hopefully continue the collaboration in the future :)

PR #44, opened 2020-09-14
Clang format lint action

marinagmoreira:
- Added github action that checks if the repo adheres to the lint check (fixed the script that checks lint errors in the entire repo)
- Added to the pre-commit an automatic clang-format run on the lines that yielded an error - this should correct straightforward errors like spacing without changing the entire file - be aware that clang-format can't correct all the errors automatically"
- Created a clang-format style definition based on google style - this might change in the future as we go.


marinagmoreira:
Github action passing:
https://github.com/marinagmoreira/astrobee/runs/1113973694

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,14 @@
+# Check for lint error and auto correct them
+
+name: Check for lint errors
+
+on: [push]
```
bcoltin, line 5: not on pull request too?

scripts/git/pre-commit
```diff
@@ -80,12 +82,27 @@ def main():
...
-        for error in output:
-            print "%s:%s: %s" % (filename, str(error[0]), error[1])
+        if len(output) > 0:
+            lines = []
+            for error in output:
+                print "%s:%s: %s" % (filename, str(error[0]), error[1])
+                lines.append(error[0])
+            # Run the clang-format if there are errors in the identified ranges
+            if len(output) > 0:
+                command = "clang-format-8 -style=file -i " + filename
```
bcoltin, line 92: How is clang-format-8 installed? Need to add to install_desktop_packages?

bcoltin (changes requested):

.github/workflows/lint.yaml
marinagmoreira: Yes that is correct, even though the user will push it to github before doing a pull request, it is useful for the reviewer to see the check. Thanks! I didn't think of that.

scripts/git/pre-commit
marinagmoreira: yes, I added clang-format-8 to the packages_desktop.lst because this is a package that does not go in the robot.

bcoltin (approved):

bcoltin:
Will wait for @rsoussan to review and then merge.

.clang-format
```diff
@@ -0,0 +1,151 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
rsoussan, line 3: Are all of these options different from Google?  If not, does it work to just override the ones that have changed?

marinagmoreira: no, just a list of all the options from:
$clang-format-8 -style=google -dump-config > .clang-format
this just dumps all the configurations available, already tuned with the google style.
It's useful to have all the configs to know exactly what are the options available if we need to change anything.
I can add in a comment on which ones were changed from the -dump-config default.

rsoussan: Ok that would be helpful

marinagmoreira: I did that in the last commit " highlight the configs that were changed from the original -dump-config in the google style"
thanks!

.clang-format
```diff
@@ -0,0 +1,151 @@
...
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     100                           # changed
```
rsoussan, line 48: I think the current linter is 120?

.clang-format
```diff
@@ -0,0 +1,151 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true              # changed
```
rsoussan, line 6: I think we disagree on this one.  I'd vote to add // clang-format off // clang-format on blocks to areas of new code where custom formatting is desirable, I'm of the opinion that the areas where we would want this are fewer than when we wouldn't, but we should probably all get a vote.  Alternatively, developers could set this as they wish, but then things could switch back and forth in the future based on who has what configs.

.clang-format
marinagmoreira: Oh that's right!

rsoussan (approved):

PR #47, opened 2020-09-16
Fix unit tests

marinagmoreira:
There are currently unit tests made with gtest and rostest. These should be run at every pull request to make sure new code does not break anything.
To do:

    Incorporate unit tests into docker action
    Go through unit tests and correct the ones that were deprecated and now fail.


marinagmoreira:
All tests pass when run natively except for test_init_mapper.
This draft pull request helps test the github action to make sure the tests pass inside docker.

marinagmoreira:
Trying to correct this error before merging in the Ubuntu 18 docker:
E0929 15:38:27.604982    94 config_reader.cc:694] ConfigReader: Unable to load libraries!! This is very bad!

.github/workflows/docs.yaml
```diff
@@ -51,4 +51,4 @@ jobs:
         git config user.email "$EMAIL"
         git config user.name "$NAME"
         git commit -m "Automatic update for $GITHUB_SHA."
-        git push origin gh-pages
\ No newline at end of file
+        { git push origin gh-pages || true; }
```
marinagmoreira, line 54: This line made the whole thing fail with "nothing to commit" when there were no changes to the gh-pages (no documentation update)

CMakeLists.txt
```diff
@@ -282,14 +282,13 @@ else (USE_CTC)
...
 # Bring in catkin
-set(CATKIN_SKIP_TESTING ON CACHE BOOL "NO")
+set(CATKIN_SKIP_TESTING OFF)
 if (USE_CTC)
   # Catkin brings in GTest as a dependency. We don't want it because
   # we have our own ... on top of that it doesn't build correctly for
   # ARM. Our own unit tests however will continue to
   # work. (non-catkin)
   set(CATKIN_ENABLE_TESTING OFF)
-  set(CATKIN_SKIP_TESTING OFF)
```
marinagmoreira, line 292: Someone should confirm this, by default we were not able to do $make tests to compile the tests because the CATKIN_SKIP_TESTING was on for native builds (not sure why this was)

astrobee/config/worlds/mgtf.config
```diff
@@ -25,8 +25,8 @@ require "rotations"
 -- Location of the dock and berth tilt values (in degrees)
 world_dock_berth1_tilt = math.rad(0)    -- Tilt in body-frame Y
 world_dock_berth2_tilt = math.rad(0)    -- Tilt in body-frame Y
-world_dock_transform = transform(0.0, 0.0, 0.0),
-  quat4(0.0, 0.0, 1.0, 1.0));
+world_dock_transform = transform(
```
marinagmoreira, line 28: This file was not formatted correctly, and the unit test failed

gnc/ctl/test/test_ctl.test
```diff
@@ -17,7 +17,6 @@
 
 <launch>
   <include file="$(find astrobee)/launch/sim.launch">
-    <arg name="gui" value="false"/>
```
marinagmoreira, line 20: we don't have this flag anymore

hardware/pico_driver/CMakeLists.txt
```diff
@@ -30,6 +30,18 @@ create_library(TARGET ${PROJECT_NAME}
...
+    test/test_init_pico_flexx.test
+    test/test_init_pico_flexx.cc
+  )
+
+  target_link_libraries(test_init_pico_flexx
+    ${catkin_LIBRARIES} config_reader ff_nodelet
+  )
+endif()
+
 create_tool_targets(DIR tools
```
marinagmoreira, line 45: I moved this test here such that it would only run if the pico_flexx compilation is present (avoiding the issue of the proprietary driver)

CMakeLists.txt
bcoltin: No idea... as long as it builds it should be fine.

scripts/docker/rebuild_test_astrobee_kinetic.Dockerfile
```diff
@@ -6,8 +6,8 @@ FROM astrobee/astrobee:latest-kinetic
...
 COPY . /src/astrobee
 RUN /src/astrobee/scripts/configure.sh -l -F -D -p /opt/astrobee -b /build/astrobee
-RUN cd /build/astrobee && make install -j`nproc`
+RUN cd /build/astrobee && make install -j6
 COPY ./astrobee/resources /opt/astrobee/share/astrobee/resources
 
 # Run tests
-RUN cd /build/astrobee && make test && make run_tests
\ No newline at end of file
+RUN cd /build/astrobee && make -j6 tests && make -j6 test
```
bcoltin, line 13: Could you explain the division between astrobee_base -> astrobee -> rebuild_test_astrobee?

Why not do astrobee (same as it used to be) -> test_astrobee? Then in test_astrobee just do make tests && make test?

marinagmoreira: I didn't think that we could chain the docker images in the github action, I'll try! That would be better yes.

marinagmoreira: Done! Thanks!

bcoltin (approved):

PR #48, opened 2020-09-17
Make astrobee media submodule

marinagmoreira:
This pull request makes astrobee_media as a shallow submodule. (easier to update)
Looks like git submodule is pretty buggy (it will not always listen to .gitmodules), so I updated the install instructions with more explanation.

### DIFF
--- a/.github/workflows/CI-build.yml
+++ b/.github/workflows/CI-build.yml
@@ -15,15 +15,21 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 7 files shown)


marinagmoreira:
The submodule points at the update with the new .ply file.
This also addresses "what to do with astrobee_android", it is clear for the user on the install instructions that it is an optional submodule.
I tested this by cloning a new repo and reproducing the commands.

bcoltin (approved):

PR #53, opened 2020-10-02
Predict function called with wrong arguments

krichardsson:
Hi!

I was looking at you implementation of the Vive localization functionality. It seems as the arguments of an atan has been reversed.

### DIFF
--- a/localization/vive_localization/src/vive.h
+++ b/localization/vive_localization/src/vive.h
@@ -265,8 +265,8 @@ static void Correct(T const* params, T * angle, bool correct) {
     ideal[0] = angle[0];
     ideal[1] = angle[1];
     for (size_t i = 0; i < 10; i++) {
-      xyz[0] = xyz[0] = T(-1.0) / tan(ideal[0]);
-      xyz[1] = xyz[0] = T(-1.0) / tan(ideal[1]);
+      xyz[0] = T(-1.0) / tan(ideal[0]);
+      xyz[1] = T(-1.0) / tan(ideal[1]);
       xyz[2] = T(1.0);
       Predict(params, xyz, pred, correct);
       ideal[0] += (angle[0] - pred[0]);
--- a/localization/vive_localization/src/vive.h
+++ b/localization/vive_localization/src/vive.h
@@ -243,7 +243,7 @@ static void Predict(T const* params, T const* xyz, T* ang, bool correct) {
       c = -xyz[0];
     }
     // Ideal angle in radians [0, 2pi] relative to the start of the sweep
-    ang[i] = atan2(a, b);
+    ang[i] = atan2(b, a);
     // Perturb this angle based on the base station parameters
     if (correct) {
       T const* p = &params[i*NUM_PARAMS];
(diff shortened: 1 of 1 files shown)


asymingt:
> Hi!
> 
> I was looking at you implementation of the Vive localization functionality. It seems as the arguments of an atan has been reversed.

I can see why you might think this to be the case, but I understand the equations as they are to be correct. This block of code is essentially calculating the sweep angle in radians with respect to the z-fwd axis (a) of the lighthouse in one of the two motor planes (b). Don't be misled by the variable xyz, which uses the astrobee z-fwd convention. This code has been tested experimentally on real data and produces the correct results.

krichardsson:
Hi Andrew and thanks for the clarification!
I was not sure how you defined the sweep angle, but now it makes sense.
My assumption came from looking at both the Predict() and Correct() functions, I believe they use different definitions of the sweep angle.
Knowing that Predict() is correct I guess the problem lies in the lines https://github.com/nasa/astrobee/blob/master/localization/vive_localization/src/vive.h#L268-L269 instead.

Changing them to
```
xyz[0] = T(-1.0) / tan(ideal[0]);
xyz[1] = T(-1.0) / tan(ideal[1]);
```
would align the definitions. Do you want me to update the pull request?

BTW we met at IROS in Madrid 2 years ago, I'm working at Bitcraze. I hope all is well!

asymingt:
> Hi Andrew and thanks for the clarification!
> I was not sure how you defined the sweep angle, but now it makes sense.
> My assumption came from looking at both the Predict() and Correct() functions, I believe they use different definitions of the sweep angle.
> Knowing that Predict() is correct I guess the problem lies in the lines https://github.com/nasa/astrobee/blob/master/localization/vive_localization/src/vive.h#L268-L269 instead.
> 
> Changing them to
> 
> ```
> xyz[0] = T(-1.0) / tan(ideal[0]);
> xyz[1] = T(-1.0) / tan(ideal[1]);
> ```
> 
> would align the definitions. Do you want me to update the pull request?
> 
> BTW we met at IROS in Madrid 2 years ago, I'm working at Bitcraze. I hope all is well!

Oh, I remember us meeting at IROS. Hello again! It's good to see you're still working with the vive system. I believe Arnaud has a working V2 base station implementation. Did you guys manage to also implement calibration / correction for this system?

With regards to your updated fix, that looks right to me. We didn't ever use `Correct` in our system, so it wasn't as well-tested as `Predict`. Are you using it on your side for stabilizing CF2s?

krichardsson:
We are still working on the vive system, but have not been able to put as mush time into it as we maybe should and unfortunately progress has been slow. Yes, Arnaud managed to implement a decoder for the V2 signals (the FPGA on the deck is stuffed!) and we have basic functionality in the CF. 

I'm having another go at the calibration/correction for V2 and was browsing libsurvive and your code for inspiration and this is when I stumbled upon this issue. I have not managed to find out how to use the curve/ogee values yet, but using only phase, tilt and gib gives a fairly good result. I've been wondering how large errors you saw when working with vive V1?

When the calibration problem is solved, what remains to implement is mainly "infrastructure" such as GUI, distributing/storing base station geometry and calibration data in a swarm and so on.

I updated the pull request (messing up the commits, sorry!). 

BTW it seems as the project is using git submodules that are not possible to clone for me (probably not public) and this causes problems when cloning and branching in git (at least from sourcetree). I had to do my changes in the github gui instead.

asymingt (approved):
Change looks good to me

bcoltin:
Thank you Kristoffer! Yes, we have some submodules we are not allowed to release.

PR #61, opened 2020-10-08
Mem monitor

marinagmoreira:
The memory monitor checks the used memory of the entire system and individual ROS nodes


scripts/build/genFaultTable.py
```diff
@@ -115,7 +115,6 @@ def main():
     file.close()
 
     faults_config_file_name = sys.argv[2] + "/faults.config"
-    faults_test_config_file_name = sys.argv[2] + "tests/faults.config"
```
marinagmoreira, line 118: I changed the astrobee/config/tests/faults.config to be a simlink of the actual astrobee/config/faults.config, therefore no need to generate it. This was done to minimize error if someone manually changes this file for testing.

astrobee/config/tests/faults.config
```diff
@@ -1,182 +0,0 @@
--- Copyright (c) 2017, United States Government, as represented by the
```
marinagmoreira, line 1: I changed the astrobee/config/tests/faults.config to be a simlink of the actual astrobee/config/faults.config, therefore no need to generate it. This was done to minimize error if someone manually changes this file for testing.

mobility/mobility/tools/teleop.cc
```diff
@@ -52,7 +52,7 @@
 DEFINE_string(ns, "", "Robot namespace");
 DEFINE_bool(reset, false, "Reset localization pipeline");
 DEFINE_bool(bias, false, "Estimate bias for the localization pipeline");
-DEFINE_string(loc, "", "Localization pipeline (none, ml, ar, hr)");
+DEFINE_string(loc, "", "Localization pipeline (no, ml, ar, hr)");
```
marinagmoreira, line 55: The actual name of the localization mode is no, just noticed this testing in the lab

bcoltin:
@marinagmoreira it looks like a test failed?

management/cpu_mem_monitor/include/cpu_mem_monitor/cpu_mem_monitor.h
```diff
@@ -0,0 +1,189 @@
...
+  // call this in regular intervals for the numbers to make sense over time.
+  int CollectCPUStats();
+  int CollectMemStats();
+
+  // Callback to scan and publish the CPU and Memory stats
+  void PublishStatsCallback(ros::TimerEvent const &te);
+
+ private:
+  struct Load {
+    std::uint64_t total_time,
```
bcoltin, line 102: I never knew you could do this! That's cool

management/cpu_mem_monitor/include/cpu_mem_monitor/cpu_mem_monitor.h
```diff
@@ -0,0 +1,189 @@
...
+  LoadFaultState load_fault_state_;
+  int assert_load_high_fault_timeout_sec_, clear_load_high_fault_timeout_sec_;
+};
+
+
+int ParseLine(char* line) {
+  // This assumes that a digit will be found and the line ends in " Kb".
+  int i = strlen(line);
+  const char* p = line;
+  while (*p < '0' || *p > '9') p++;
```
bcoltin, line 167: add check for string length

management/cpu_mem_monitor/src/cpu_mem_monitor.cc
```diff
@@ -0,0 +1,665 @@
...
+  std::map<std::string, int>::iterator it;
+  for ( it = nodes_pid_.begin(); it != nodes_pid_.end(); it++ ) {
+    // Look for PID if not already on the list
+    if (it->second == 0) {
+      // Check if the node is being executed in this computer
+      // Get URI of the node
+      args.setSize(2);
+      args[0] = ros::this_node::getName();
+      args[1] = it->first;
+      ros::master::execute("lookupNode", args, result, payload, true);
```
bcoltin, line 308: check return code

management/cpu_mem_monitor/src/cpu_mem_monitor.cc
```diff
@@ -0,0 +1,665 @@
...
+  }
+
+  for (unsigned int i = 0; i < load_cpus_.size(); i++) {
+    if (fgets(buffer, sizeof(buffer), statf) == NULL) {
+      fprintf(stderr, "fgets screwed up\n");
+      break;
+    }
+
+    if (i == 0) {
+      sscanf(buffer, "cpu  %" SCNu64 " %" SCNu64 " %" SCNu64 " %" SCNu64 \
```
bcoltin, line 379: check return codes from sscanf

management/cpu_mem_monitor/src/cpu_mem_monitor.cc
```diff
@@ -0,0 +1,665 @@
...
+      temp_fault_triggered_ = false;
+    }
+  }
+}
+
+int CpuMemMonitor::CollectMemStats() {
+  // Update memory info
+  // In the mem_info_ structure, sizes of the memory and swap
+  // fields  are  given  as  multiples  of mem_unit bytes.
+  sysinfo(&mem_info_);
```
bcoltin, line 534: check return code

bcoltin (changes requested):
Just a couple very minor things and let's fix the unit tests, then I think we are good to go!

management/cpu_mem_monitor/include/cpu_mem_monitor/cpu_mem_monitor.h
```diff
@@ -0,0 +1,189 @@
...
+  LoadFaultState load_fault_state_;
+  int assert_load_high_fault_timeout_sec_, clear_load_high_fault_timeout_sec_;
+};
+
+
+int ParseLine(char* line) {
+  // This assumes that a digit will be found and the line ends in " Kb".
+  int i = strlen(line);
+  const char* p = line;
+  for (uint j = 0; ( *p < '0' || *p > '9' ) && j < strlen(line); ++p, ++j)
```
marinagmoreira, line 167: I didn't know you could add multiple lines to the for loop!

bcoltin: This is going to execute line[i-3] = '\0'; once each time through the for loop, which won't break anything but probably isn't what you want. Also doesn't matter but you could have also stuck to the while loop and done p - line < i

marinagmoreira: totally forgot the ;! That would have been a slick solution, I didn't do that because p is a character, so I introduced a counter.

chloekek: You could remove the `j` variable and simplify the search like this, as strings always end with `'\0'` (in fact, `'\0'` is what `strlen` itself looks for):

```cpp
for (; (*p < '0' || *p > '9') && *p != '\0'; ++p) {}
```

Or use `std::find_if` to get a pointer to the first digit, and have it handle the boundary conditions:

```cpp
const char* p = std::find_if(line, line + i,
                             [] (char c) { return c >= '0' && c <= '9'; });
```

bcoltin (approved):

PR #63, opened 2020-10-21
Review fligh mode parameters

marinagmoreira:
replace extra flight modes with precision
change speed for 2


marinagmoreira:
Awaiting test in the granite lab

marinagmoreira:
I tested this in the lab, basically replaced all of the special behaviors flight modes with a PRECISION to be easier to manage. The PRECISION is the same as nominal, but with fan speed of 3

bcoltin (approved):

bcoltin:
@kbrowne15 can you double check this and then we can merge

kbrowne15 (approved):

PR #64, opened 2020-10-21
Add fault documentation

kbrowne15:
Moved the FMECA spreadsheet from freeflyer_docs to the doc folder. Also added documentation on how to add faults and modified the generate faults config script to account for the FMECA being moved.


bcoltin:
It looks like it is generated from the csv file... can we not just save the csv file in the repo? Having code generated from Excel spreadsheets terrifies me

doc/general_documentation/adding_a_fault.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Fault Config Files Generation
+
+There is a script that will generate the fault config files by parsing the FMECA spreadsheet. This script assumes the steps in the previous section were completed and the FMECA spreadsheet was saved as a csv file in the doc folder. If it was not saved in the doc folder, please set the FMECA_CSV_PATH environment variable to folder were it was saved. Then run the following line:
```
marinagmoreira, line 13: If I save as csv, it only saves the first page, is that ok? (skips the not monitored faults)

marinagmoreira (commented):
Given that the table is really big, having it directly in csv would be complete gibberish (it would also lose information so a different file would need to be kept separately to keep track of it). Maybe an alternative would be json/yaml?

kbrowne15:
We have decided that the FMECA should be converted to a json file much like our command schema. The tools that process the FMECA spreadsheet will also need to be converted.

bcoltin:
Katie would still like to do this, and may have time early in 2024. If not we should close this.

PR #67, opened 2020-10-27
Cross compile docker

marinagmoreira:
Cross compiler docker for astrobee
This takes into assumption that the rootfs folder was already generated and the environment variables defines (added to the readme instructions)


scripts/docker/cross_compile/cross_compile.sh
```diff
@@ -0,0 +1,34 @@
...
+# This image is split into two, because we need to copy over the toolchain
+# folder and rootfs folder. Because docker can only access files in the
+# build context, and we don't want it to scan the entire computer by
+# chosing a bigger build context, this is the easiest solution. This also
+# allows the user to place the folders freely.
+
+# Base docker image copies over the toolchain and rootfs, resulting from
+# the NASA_INSTALL.md instructions. Also installs minimum dependencies
+echo "Build context for base: "${ARMHF_CHROOT_DIR/../}
+docker build $ARMHF_CHROOT_DIR/../ -f scripts/docker/cross_compile/astrobee_base_cross.Dockerfile -t astrobee/astrobee:base-cross
```
bcoltin, line 29: In other scripts we use the ARMHF_TOOLCHAIN variable and these two don't have to be in the same place... could we make the docker respect this as well?

bcoltin: also we are assuming that the chroot dir is called ${ARMHF_CHROOT_DIR}/../arm_cross which the other tools do not assume

bcoltin (changes requested):

scripts/docker/cross_compile/cross_compile.sh
bcoltin: I guess this would make us have two base files with two contexts. Another option could be we would have one environment variable for both of these and a script / instructions to check them out in the right way, and change everything else to only use this variable.

marinagmoreira: The variables are extensively used throughout the code, it makes sense to keep them separate for readability. I'll just add another build context.

marinagmoreira: I've corrected this :) Thank you!!

scripts/docker/cross_compile/cross_compile.sh
```diff
@@ -25,8 +25,11 @@
 
 # Base docker image copies over the toolchain and rootfs, resulting from
 # the NASA_INSTALL.md instructions. Also installs minimum dependencies
-echo "Build context for base: "${ARMHF_CHROOT_DIR/../}
-docker build $ARMHF_CHROOT_DIR/../ -f scripts/docker/cross_compile/astrobee_base_cross.Dockerfile -t astrobee/astrobee:base-cross
+# echo "Build context for base: "${ARMHF_TOOLCHAIN}
```
bcoltin, line 28: It looks like they are commented out?

marinagmoreira: bah forgot to uncomment after doing some tests, fixed

bcoltin: the code review is working :)

bcoltin (approved):

PR #71, opened 2020-11-04
Env overlay correction

marinagmoreira:
To get the unit tests working I had to switch the order of the catkin_package declaration in the cmake files, turns out that this interferes with the environment overlay for isaac. The problematic tests were converted to rostest (still gtest) to solve the problem and allow correct order in the cmake.


localization/camera/CMakeLists.txt
```diff
@@ -46,16 +53,26 @@ create_tool_targets(DIR tools
...
-)
+  add_rostest_gtest(test_camera_model
+    test/test_camera_model.test
+    test/test_camera_model.cc
+  )
+  target_link_libraries(test_camera_model
+    camera
+  )
+
+  find_package(rostest REQUIRED)
```
bcoltin, line 67: Why are there multiple find_package rostests? Our standard has been to import most things in the top level CMakeLists.txt, could we move this there?

localization/interest_point/CMakeLists.txt
```diff
@@ -50,17 +57,19 @@ create_tool_targets(DIR tools
...
-)
 
-catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES interest_point
-  CATKIN_DEPENDS
-#  DEPENDS system_lib
-)
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
```
bcoltin, line 62: same

localization/sparse_mapping/CMakeLists.txt
```diff
@@ -60,16 +66,54 @@ if (NOT USE_CTC)
     EXCLUDE ${EXCLUDE})
 endif (NOT USE_CTC)
 
-create_test_targets(DIR test
-  LIBS ${LIBS} sparse_mapping
-  INC ${INCLUDES}
-  DEPS ${DEPS})
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
```
bcoltin, line 70: same

bcoltin: lots of these in this file?

shared/ff_common/CMakeLists.txt
```diff
@@ -17,20 +17,29 @@
...
 create_library(TARGET ff_common
   INC ${GFLAGS_INCLUDE_DIRS} ${GLOG_INCLUDE_DIRS}
   ${CMAKE_CURRENT_BINARY_DIR}/include
   LIBS ${GFLAGS_LIBRARIES} ${GLOG_LIBRARIES})
 
-create_test_targets(DIR test
-  LIBS ff_common
-  DEPS ff_common)
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
```
bcoltin, line 32: same here

bcoltin (changes requested):
Can we move "find_package(rostest REQUIRED)" into top CMakeLists.txt and only call it once? Then I think it is good.

localization/camera/CMakeLists.txt
marinagmoreira: Thanks for noticing the repeated includes! Definitely a copy paste error.
I don't think we have the package rostest when cross-compiling, that's why it is inside the CATKIN_ENABLE_TESTING flag. This is a convention across all unit tests in the repo.

bcoltin (approved):

PR #78, opened 2020-11-18
Behaviour state machine review

marinagmoreira:
Things done:
- Improved feedback
-Arm will go to 0,0 if deploy command
-Perching status had repeated state, and made the 0-UNPERCHED, 1-PERCHED in case we want to manually set it to be easier


communications/ff_msgs/msg/PerchState.msg
```diff
@@ -23,33 +23,33 @@ std_msgs/Header header
...
+int8 PERCHING_ENSURING_APPROACH_POSE    = 10
+int8 PERCHING_DEPLOYING_ARM             = 9
+int8 PERCHING_OPENING_GRIPPER           = 8
+int8 PERCHING_MOVING_TO_COMPLETE_POSE   = 7
+int8 PERCHING_CLOSING_GRIPPER           = 6
+int8 PERCHING_CHECKING_ATTACHED         = 5
+int8 PERCHING_WAITING_FOR_SPIN_DOWN     = 4
+int8 PERCHING_SWITCHING_TO_PL_LOC       = 3
+int8 PERCHING_STOPPING                  = 2
+int8 PERCHED                            = 1
```
kbrowne15, line 45: These values are based on the mobility state that Lorenzo defined a long time ago and docking matches. Basically, we have N to 1 as perching, 0 as perched, and -1 to -N as unperching. If you change unperch to 0, the executive/GDS will display that Astrobee is perched when Astrobee is unperched. If there is a reason unperched needs to be 0 then I will have to change the executive to map the values correctly. If there isn't a reason, please change them back to what they were.

kbrowne15 (changes requested):

communications/ff_msgs/msg/PerchState.msg
marinagmoreira: Thank you! I didn't know about that convention, then unperch can be -1, and still something easy to remember

communications/ff_msgs/msg/PerchState.msg
```diff
@@ -43,13 +44,14 @@ int8 PERCHING_WAITING_FOR_SPIN_DOWN     = 3
...
 int8 PERCHING_STOPPING                  = 1
 int8 PERCHED                            = 0
-int8 UNPERCHING_SWITCHING_TO_HR_LOC     = -1
-int8 UNPERCHING_WAITING_FOR_SPIN_UP     = -2
-int8 UNPERCHING_OPENING_GRIPPER         = -3
-int8 UNPERCHING_MOVING_TO_APPROACH_POSE = -4
-int8 UNPERCHING_STOWING_ARM             = -5
-int8 UNPERCHING_SWITCHING_TO_ML_LOC     = -6
-int8 UNPERCHED                          = -7
+int8 UNPERCHED                          = -1
```
kbrowne15, line 47: Unfortunately, this still breaks some assumptions in the executive and GDS. I think this may just have to be something an operator looks up.

kbrowne15 (changes requested):

communications/ff_msgs/msg/PerchState.msg
marinagmoreira: Alright then :+1:

kbrowne15 (approved):

PR #84, opened 2020-11-26
I added some minor functionality to gds_simulator.py and some doc tweaks

oleg-alexandrov:


tools/gds_helper/src/gds_simulator.py
```diff
@@ -363,12 +364,21 @@ def execute_action(selection):
             print_app_cmd(selection)
             print str(num_cmds + 1) + ')  Exit program'
             try:
-                command = input('\nType the number of the command you wish to send: ')
+                command = raw_input('\nType the number of the command you wish to send or the command string: ')
+                if len(command) >= 1 and command[0] == '{':
```
kbrowne15, line 368: I'm not sure we can assume that a guest science command starts with a '{'. We suggest the custom guest science commands are json formatted but they don't have to be. Perhaps a better check would be to see if command[0] is not numeric.

kbrowne15 (changes requested):

tools/gds_helper/src/gds_simulator.py
oleg-alexandrov: I put a fix to check if the command is numeric and tested it on the bot. The commit is below.

marinagmoreira (approved):

kbrowne15 (approved):

PR #90, opened 2020-12-11
Allow image undistortion at resolution differing from native. Doc tweaks

oleg-alexandrov:
This is some minor functionality needed for ISAAC which does not break anything.

### DIFF
--- a/localization/camera/include/camera/camera_params.h
+++ b/localization/camera/include/camera/camera_params.h
@@ -71,7 +71,7 @@ namespace camera {
     // Used to create a remap table. This table is the same size as the
     // UNDISTORTED image, where every pixel's value is the cooresponding pixel
     // location in the DISTORTED image.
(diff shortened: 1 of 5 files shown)


marinagmoreira:
Seems to fail for Ubuntu 18 with:
/src/astrobee/localization/camera/tools/undistort_image.cc: In function 'int main(int, char**)':
/src/astrobee/localization/camera/tools/undistort_image.cc:139:37: error: no match for 'operator*' (operand types are 'double' and 'const Vector2i {aka const Eigen::Matrix<int, 2, 1>}')
   Eigen::Vector2i dims = FLAGS_scale*cam_params.GetDistortedSize();

oleg-alexandrov:
Thank you. I put a fix. I will keep an eye on how the compilation goes.

oleg-alexandrov:
I think it worked this time.

marinagmoreira (approved):

PR #96, opened 2021-01-05
Add pictures of where granite lab registration corners are

oleg-alexandrov:
Also refactor the doc so that it looks good on the web.


bcoltin (commented):
The images are a bit big, 2 MB total, which I know isn't much but it adds up. Could we convert to jpg? And possibly downsample if there is a level that you still think looks nice

oleg-alexandrov:
I made them jpeg. The size went down a lot.

bcoltin:
Perfect, thank you!

PR #97, opened 2021-01-05
Added flight release documentation.

kbrowne15:
Added the flight release documentation. It may need minor tweaks but I wanted to get everyone's comments sooner than later.


doc/general_documentation/flight_release.md
```diff
@@ -0,0 +1,89 @@
...
+Please make sure to commit and push all the files that were changed.
+
+    git commit -m "meaninful message" files_that_changed
+    git push -u origin release-0.x.x
+
+
+## Create the Debian package
+
+
+    ./scripts/build/build_debian.sh
```
bcoltin, line 45: Mention that cross compile needs to have been setup?

doc/general_documentation/flight_release.md
```diff
@@ -0,0 +1,89 @@
...
+    git checkout -b release-0.x.x
+
+
+## Update the Release Version
+
+The release version must be updated before creating the Debian. Please use
+the update release script to do this. Before running the script, get a list of
+features that have been added since the last release. This list will need to be
+added to the change log that gets edited when the script is run. Also, in the
+change log, change UNRELEASED to testing.
```
bcoltin, line 26: and change the email address to the FSW email address in the previous log entries

bcoltin (changes requested):
A couple minor comments but mostly looks good!

doc/general_documentation/flight_release.md
kbrowne15: I did at the beginning but I'll move it down to this section.

doc/general_documentation/flight_release.md
```diff
@@ -0,0 +1,89 @@
+\page release Creating a Flight Release
+
+# Creating a Flight Release
+
+This page describes the steps required to complete a flight release. It assumes
+that your environment is configured to cross-compile the fsw for the `armhf`
+architecture.
+
+## Create a Release Branch
```
marinagmoreira, line 9: Could we use the github release feature instead or in addition to this? 
https://github.com/nasa/astrobee/releases
That becomes more accessible for outside users and all the versions become easily tractable on the releases page.

marinagmoreira: Actually this shouldn't be here, but in the end on the "reviewers" section after everything is tested

trey0 (changes requested):
Thanks, overall this looks great and it's a big step forward for better CM. Just a couple of comments below.

doc/general_documentation/flight_release.md
bcoltin: I like it! (and this will also create a tag)

doc/general_documentation/flight_release.md
```diff
@@ -0,0 +1,89 @@
...
+
+Release branches will be created on a user's fork. To make a release branch,
+checkout and update develop. Then make a release branch off of develop.
+
+    git checkout develop
+    git fetch upstream
+    git merge upstream/develop
+    git checkout -b release-0.x.x
```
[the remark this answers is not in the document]
bcoltin: The debians have a git hash embedded in them which is already reported by the fingerprinting script. So I think we should not bump the version number unless the version has already been accepted and merged into develop / master.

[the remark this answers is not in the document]
trey0: Great! I like that approach. I recommend documenting it with something like the following:

> Note that creating a release branch and specifying a version number indicates that the code on that branch is frozen with respect to further feature changes. However, as you identify bugs during lab testing before deployment to the ISS, you can still apply bug fixes on that release branch without incrementing the version number. Of course, that means that as you iterate, you might end up generating multiple debs with the same version number. For configuration management purposes, when checking whether two deployments used identical debs, you should always compare their git commit hashes rather than their version numbers.

marinagmoreira (approved):

doc/general_documentation/flight_release.md
```diff
@@ -0,0 +1,99 @@
...
+
+On confluence, there is a [release testing procedure](https://babelfish.arc.nasa.gov/confluence/display/FFFSW/IRG-FFTEST302+-+Astrobee+Release+Testing).
+This procedure will walk you through all the testing that needs to be done to
+make sure the debian is flight ready. Please create a copy of the procedure by
+clicking on the ellipsis at the top right corner of the page and selecting copy.
+In the popup, in the 'Parent page' text box, please enter 'Executed Procedure'
+and click 'Copy'. Remove 'Copy of' from the title and add the version and date
+after the test number such that it looks like
+'IRG-FFTEST302 - 0.x.x -YYYY-MM-DD - Astrobee Release Testing'. Be sure to fill
+out the 'CAs-Ran Summary' table and the 'Result' and 'Notes' column for every
```
trey0, line 65: Typo 'CAs-Ran ...'

trey0 (approved):
Thanks! All of my previous comments were addressed.

doc/general_documentation/flight_release.md
kbrowne15: Thanks for catching that!

bcoltin (approved):

bcoltin:
@kbrowne15 I think you got everything but not entirely sure, can you merge it when it's ready? Thanks

PR #106, opened 2021-01-21
Make debian docker

marinagmoreira:
Adding a docker image that generated the debians and a script that runs everything and copies the files out.
Took out some useless steps on the cross compile docker.


marinagmoreira:
I added both Brian and Katie as reviewers, but only one needs to do it, I just wasn't sure who to ask.

bcoltin (approved):

PR #110, opened 2021-01-27
Feature/aahs 513 r2 astrobee simulation

nicole-j-ortega:
adding new world to astrobee that enables R2 launch files to take care of spawning R2 and the US lab
** I understand that astrobee_media needs to be merged first, which will require reverting the state of .gitmodules


trey0:
It's new to me that we would want to modify the main astrobee repo in order to support integrated simulation with R2. I had initially imagined that we would be able to create some glue code that would reference and integrate the simulator environments of both robots, with that glue code living in a separate repo. But I can see why some changes in the astrobee repo might be needed.

In that case, my tentative opinion is any code in the astrobee repo that explicitly references R2 is probably too application-specific to be merged with the astrobee 'develop' branch. Logically, it could either live on an integrated-sim-specific branch, or perhaps it could be refactored so that instead of referencing R2, it creates a more generic hook that satisfies the needs of the integrated sim, but could also be repurposed by other users.

Some other considerations to point out: we don't want to accidentally release any R2-related information that is not cleared for open-source release, nor accidentally create dependencies of Astrobee FSW on R2 software. Maybe that's already handled properly in this PR, just wanted to point out it could be an ongoing issue.

(All of that said, I want to defer to Brian and Katie's opinion.)

kbrowne15:
I agree with Trey. I really don't think it should be merged into the astrobee develop branch. However, we are planning on creating a branch that will be ISAAC specific so it may make sense to add it to that branch. If we do decide to go that way, I want to reiterate what Trey said, we will definitely have to make sure we don't accidentally release anything R2 related that shouldn't be released.

nicole-j-ortega:
There are no R2 specific changes in this merge request. I intentionally created changes that are completely independent from R2 and could be repurposed by others. The only change made is creating a "new" world that has the US module removed from the ISS. This allows R2 and the US module to spawn independently. For grasping purposes, the US lab must include rails that use a gazebo plugin specific to R2.

To eliminate any relevance to R2, I can simply change the name of this new world from R2 to something else.
________________________________

bcoltin:
Hi Nicole,

Copying the ISS world makes it more difficult maintain. Would it be possible to conditionally add or not add the US module using xacro instead?

Also, if this is just removing the US module from the world, what are the new R2 files in astrobee_media?

nicole-j-ortega:
Hi Brian,

That sounds like a better approach. However, Marina was thinking of adding these changes into the ISAAC repo instead and requested I keep this MR open for her. I believe we were going to close this request once she merged her changes to the ISAAC repo. Sorry for the miscommunication.

-Nicole

marinagmoreira:
Closing this to integrate it in isaac instead

PR #115, opened 2021-02-03
Use sim time check ekf

marinagmoreira:
adding a check to be able to use sim time, and have the simulation slower than real time

### DIFF
--- a/gnc/ekf/src/ekf_wrapper.cc
+++ b/gnc/ekf/src/ekf_wrapper.cc
@@ -31,7 +31,7 @@
 namespace ekf {
 
 EkfWrapper::EkfWrapper(ros::NodeHandle* nh, std::string const& platform_name) :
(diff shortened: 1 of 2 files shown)


gnc/ekf/src/ekf_wrapper.cc
```diff
@@ -368,7 +370,7 @@ int EkfWrapper::Step() {
     if (!have_imu_) {
       imus_dropped_++;
       // publish a failure if we stop getting imu messages
-      if (imus_dropped_ > 10 && ekf_initialized_) {
+      if (imus_dropped_ > 10 && ekf_initialized_ && !sim_) {
```
bcoltin, line 373: May be a good idea to add a comment to explain this

marinagmoreira:
This bug is not present on the new release 0.14.0, closing this.

PR #122, opened 2021-02-08
Merging master back into develop

marinagmoreira:


bcoltin:
Not sure why github was showing files changed here. I tried doing the merge from the command line and on the command line it showed no changes, so I pushed that. Then it closed this pull request so it must have done the same thing. Maybe we found an edge case in github's "Files changed" feature.

Issue #131, opened 2021-02-12
Change docking behavior


kbrowne15:
Docking behavior always switches to sparse mapping at the beginning of a dock action. Change this to check if we are in ar tag localization and to not switch to mapped landmarks if we are.

PR #133, opened 2021-02-18
Don't swich to ML if in AR loc dock behavior 

marinagmoreira:
This addresses the issue #131
I've tested this in sim, and will test again in the Lab (2/19/2021)

Note: for some reason, github is tracking changes in  astrobee/scripts/docker/build.sh, but if we do diff from the cmd line there are no changes.


bcoltin:
Looks good, I will leave this for you to merge after you test. Very weird with github's diff, not sure what is going on...

marinagmoreira:
Just tested it in the lab, everything looks good, merging

PR #135, opened 2021-02-25
Add dynamic imu filtering, better imu configs, option for optical flow measurement spacing, and small fixes

rsoussan:
The number of files changed in this pr is quite large, but I had to fix a bug where the localizer wasn't using the correct imu filters so I think the changes are pretty important.  In addition to that, there are some config and script changes, a change for optical flow measurement spacing (adding an option to increase the measurement spacing), and a change to add the option to limit the number of localization factors added to the graph so it doesn't get overloaded by too many map matches, a fix so that the reported feature counts in the localizer are more in line with what the facilities team expects, and more introspection added to the graph msg. I may update some of the scripts and configs before Friday, but ideally the runtime code will remain the same.  Thank you


bcoltin (approved):
Looks good, but there are a lot of changes involving the fan speed. Make sure you test these in granite lab!

marinagmoreira (approved):

marinagmoreira:
Let's do some tests on Monday before merging and starting the release procedure.

PR #141, opened 2021-03-05
Add calibration for haz to nav and sci cam extrinsics, and add a warning for loading very old maps

oleg-alexandrov:


bcoltin (approved):

bcoltin:
@marinagmoreira Could you merge this after the code freeze is over? (I think after we merge back from master to develop it is done, right?)

PR #149, opened 2021-03-11
Fix platform name + better default_robot naming

marinagmoreira:
platform name checked after init; adding / line to make it work
changing default to default_robot for better understandability

ps. not everyone needs to review this, just whoever has time

### DIFF
--- a/astrobee/launch/sim.launch
+++ b/astrobee/launch/sim.launch
@@ -50,7 +50,7 @@
   <arg name="stats" default="false" />          <!-- Message statistics      -->
 
   <!-- Debug-specific options -->
-  <arg name="default" default="true" />         <!-- Insert default robot    -->
+  <arg name="default_robot" default="true" />   <!-- Insert default robot    -->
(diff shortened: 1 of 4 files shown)


rsoussan:
Looks good - I think you need to do the same change for graph_localizer_nodelet.cc as done for imu_augmentor_nodelet.cc

oleg-alexandrov:
Marina, thank you for renaming 'default' to 'default_robot'. I think this should be done also in isaac, for consistency, as the same logic is used there. Hope it won't be too much extra work.

bcoltin (changes requested):
Let's merge in develop to pass the linter and respond to Ryan's suggestion and then we should be good.

marinagmoreira:
> Looks good - I think you need to do the same change for graph_localizer_nodelet.cc as done for imu_augmentor_nodelet.cc

Good point, does the imu_aug need:
ff_common::InitFreeFlyerApplication(getMyArgv());
there too?

bcoltin (approved):

PR #159, opened 2021-03-30
Added the generic action.

kbrowne15:


kbrowne15:
We are going to do this a different way.

PR #163, opened 2021-04-01
Extended the position and attitude tolerances

marinagmoreira:
extended the tolerances for the nominal mode, both for the granite lab and the iss worlds
Hopefully this will trigger the anomaly less, that completely stops the commanded movement


bcoltin (commented):
Do we think 20 cm will be enough? From the bags that failed before, would this have been sufficient? I'm wondering if we might need to make it even bigger.

bcoltin (approved):
Thanks!

PR #166, opened 2021-04-05
Updated Smart Factor Spacing and Added Separation Criteria

rsoussan:
This fixes some of the issues we've seen in CM11 and ISAAC tests, primarily a bug where we were iterating the wrong way when adding smart factor measurements and adding the oldest and not the most recent.  This caused an issue where sometimes we received a loc measurement and it was attached to a node without an optical flow constraint and caused odd velocity changes as a result.  Additionally, this led to longer imu extrapolation than desired.  
I've also added some evaluation scripts and modified them to allow for comparing with a separate groundtruth bag.


localization/graph_localizer/src/feature_tracker.cc
```diff
@@ -24,96 +24,106 @@ namespace lc = localization_common;
...
 boost::optional<lc::Time> FeatureTracker::PreviousTimestamp() const {
-  for (const auto& feature_track_pair : feature_tracks_) {
-    const auto& points = feature_track_pair.second.points;
-    if (points.size() < 2) continue;
-    return points[points.size() - 2].timestamp;
-  }
-  return boost::none;
+  const auto& longest_feature_track = LongestFeatureTrack();
+  if (!longest_feature_track) return boost::none;
+  // TODO(rsoussan): Need to check this before returning? If boost::none, is this cast correctly when returned here?
```
bcoltin, line 115: TODO

bcoltin (approved):

bcoltin:
@rsoussan I think it's fine to fix it but I'll let you merge it in case you want to do something about the minor things I mentioned or anything else before tomorrow.

bcoltin:
*fine to merge it

marinagmoreira (approved):

PR #170, opened 2021-04-08
Allowing to update the PMC Control minimum Rate

Pedro-Roque:
Simple commit but that can save bandwidth when using controllers that do not require high control rates. Saves system resources by avoiding the repetition of equal messages to meet the default 62.5 PMC frequency.

By changing the minimum control rate, the last received message is kept and used for controlling the Astrobee. An added parameters, control_min_rate_hz_, that can be adjusted by the Astrobee team to a "safe" value, below which frequencies are not accepted. 

Tagging @bcoltin and @marinagmoreira . Also, @bcoltin , added a TODO on the meaning of the 20.0 on the ros::Duration for the watchdog timer.

@albee


bcoltin:
Pretty sure 20 is just a value that worked for us, detected failures reasonably quickly without giving us false positives. What control rate are you all using? If you are doing something much slower, i.e., the 2 s, this would mean you wouldn't detect your controller stopped giving inputs for 20s. (where did the 0.5 minimum come from?) Regardless, if minimum control is is n, I would suggest setting the timer to something like min(20 / rate, 2 * n) might be better. What do you think?

Pedro-Roque:
> Pretty sure 20 is just a value that worked for us, detected failures reasonably quickly without giving us false positives. What control rate are you all using? If you are doing something much slower, i.e., the 2 s, this would mean you wouldn't detect your controller stopped giving inputs for 20s. (where did the 0.5 minimum come from?) Regardless, if minimum control is is n, I would suggest setting the timer to something like min(20 / rate, 2 * n) might be better. What do you think?

The 0,5Hz is a value that maybe you folks can adjust to something that makes more sense - it's simply a minimum rate bound, so that people cannot just put a very large cutoff time. And then I also don't allow for a smaller cutoff time (since it might lead to instability issues or Zeno).

I could implement min(20 / rate, 2 * n), but we would loose the maximum bound.

Regarding the number 20, got it! I thought it was some wierd hardware constant, but in that case I will remove it! 

Thanks @bcoltin !

Pedro-Roque:
Just pushed an update. Maybe for clarity, we should define a "safe_rate" parameter that is used inside the PMC nodelet, with value 0.05*control_rate_hz ? 

Pedro,

bcoltin:
Hi Pedro,

Sorry, I didn't explain this well. The watchdog timer is checking to see if a command has been received in the last watchdog_period_ time period. If it doesn't it will shut down and idle the propulsion. This is a quite drastic action and we want some leeway. So I would keep the factor of 20. Using just 1/rate will definitely not work and will cause propulsion to constantly idle. Using the  min(20 / rate, 2 * n) would set a maximum bound of 1 s... So even if the rate is 0.5 Hz, we would require an update every 1s instead of every 10s. It also has the advantage of not changing the current behavior for the current rate, which we haven't had issues with.

Can you please change this, and also fix the linter error? Thank you!

Pedro-Roque:
> Hi Pedro,
> 
> Sorry, I didn't explain this well. The watchdog timer is checking to see if a command has been received in the last watchdog_period_ time period. If it doesn't it will shut down and idle the propulsion. This is a quite drastic action and we want some leeway. So I would keep the factor of 20. Using just 1/rate will definitely not work and will cause propulsion to constantly idle. Using the min(20 / rate, 2 * n) would set a maximum bound of 1 s... So even if the rate is 0.5 Hz, we would require an update every 1s instead of every 10s. It also has the advantage of not changing the current behavior for the current rate, which we haven't had issues with.
> 
> Can you please change this, and also fix the linter error? Thank you!

Yep will do! Yes, I got the FAM part, but maybe I didn't explain myself well ahah. The "safe_rate" - that maybe I will change the name to "max_timeout" (since we don't care about rates here anyway)  - is just to limit how large people can change that idling. Perhaps we don't want a Astrobee that is allowed to have an input only at every 10 seconds (10 seconds being exageratted, but lets say 2 seconds), since it might hit walls and such...

So I will:
- Keep the constant you have, the 20/control_rate_hz, since that you will never change;
- Change the safe_rate to max_timeout, and the user specifies the timeout in seconds instead of having to convert to rates;
- Fix the linting error.

I hope this is what we both meant :smile:

bcoltin:
I think that sounds right! Thank you!

Pedro-Roque:
@bcoltin should be good to go

bcoltin (approved):

bcoltin:
@Pedro-Roque Actually, sorry, I had another thought after I saw your changes. Is there a need to change this value during execution? What if we just loaded the timeout from the config file at the start and dispensed with the service entirely? If there is a need for changing the timeout at runtime though this change is fine.

Pedro-Roque:
@bcoltin on my case I plan to have routines with MPC and routines with the onboard PID (mostly going back to origin after having atomic tests of mobility and state estimation). For maneuvers using PID, I'd be more comfortable setting the default value for cutoff, while for my maneuvers I believe having a larger timeout is better. 

If think its better to keep a desired cutoff on the config file (independent of the control frequency so that its clear its meaning), then I can do that change too (would remove the added service and just load a different parameter so that the watchdog timer is "pmc_timeout" instead of "20.0/control_rate_hz_", so others can modify it too).

Let me know what you think is best. If you don't think it makes sense to change it mid-testing, then we can keep the easy route.

albee:
@bcoltin Just wanted to add that the way we've previously gotten around the 62.5 Hz requirement when our controller is running at a slower rate is to do a zero-order hold of the last control input sent out at 62.5 Hz and updated as soon as new control values are available. We've typically run MPC at ~5 Hz before for example. This feature would be useful so that we don't have to set up that separate 62.5 Hz publisher. I would definitely recommend a sanity check on the hardware though, since AFAIK the impeller shutdown does not happen in sim.

bcoltin:
Ok sounds good, if it's useful we will merge it in then. Thanks!

PR #171, opened 2021-04-15
Reverted Smart Factor Spacing to 14.1, Added relative rmse comparison and Imu bias tester adder

rsoussan:
Reverted smart factor spacing as the previous version was more robust.  Added option so this can still be switched to new spacing.  Also added support in scripts for relative rmse comparison and a tool to add imu bias estimates to an existing bag file.


bcoltin (approved):

bcoltin:
Fix the lint errors and then it looks good... are we testing today?

bcoltin:
@rsoussan oops... did we forget to merge this? Is this still good to merge?

rsoussan:
Oh yeah this should be merged

PR #172, opened 2021-04-21
Add imu bias config to sim

rsoussan:
Also added some logic to ground truth localizer so that it responds to service commands even if they don't serve a purpose in ground truth mode (otherwise the loc manager waits for a response and stalls).

### DIFF
--- a/astrobee/launch/sim.launch
+++ b/astrobee/launch/sim.launch
@@ -221,4 +221,7 @@
     </include>
   </group>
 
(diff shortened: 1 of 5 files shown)


rsoussan:
@KhaledSharif this fixes the imu bias initialization in simulation we had chatted about a month or so ago

bcoltin (approved):

marinagmoreira (approved):

PR #174, opened 2021-05-05
Update bumble nav to sci and haz cam extrinsics

oleg-alexandrov:
This is a minor update which will only affect ISAAC. It makes small improvements to the transforms from the nav camera to the sci and haz ones. 

I don't understand why GitHub keeps on insisting that my pull request has 10 commits even though I changed just one file and my git branch is synched up with the develop branch.


trey0:
> I don't understand why GitHub keeps on insisting that my pull request has 10 commits even though I changed just one file and my git branch is synched up with the develop branch.

It looks like you merged changes from astrobee:develop into oleg-alexandrov:develop. That means that the two develop branches have different history in terms of merge commits, which propagates to other commits since then having different ids so they show up in the PR.

If you would like to have a clean commit history showing up in your PR (I agree it's nice), recommended workflow steps are:
- Instead of always working on an oleg-alexandrov:develop branch, try to segment your work into feature branches that correspond to the PRs you expect to make. Start each new feature branch from astrobee:develop. That should mostly fix the problem by itself.
- If some other changes are merged into astrobee:develop while you work, you can pull those changes into your feature branch using 'git rebase' instead of 'git merge', to avoid having different merge commits. (But be careful using 'git rebase' if you are already collaborating on the branch with other developers, because it rewrites history in a way that can cause confusion. In that case you may prefer to create a brand-new branch based on your feature branch and do the rebase and PR on the new branch.)

oleg-alexandrov:
Thank you. That is very helpful. And yeah, 'git rebase' is evil, unlike what Zack used to think.

bcoltin (approved):

bcoltin:
Another thing to keep in mind is, we squash and merge all the pull requests, so it doesn't really matter what the history of the PR is once it gets merged into develop

PR #176, opened 2021-05-07
Added re-plan command to the executive

kbrowne15:


marinagmoreira (approved):

trey0:
This seems like a great idea to me! However, I would like to help make the API documentation more specific, particularly thinking about novice guest scientist users who might have questions about the semantics.

Below I took a stab at editing the descriptions for specificity, but please sanity check if I'm describing the intended behavior correctly! Back when Andrew was around we had a lot of talk about cool features that he wanted to eventually implement in the choreographer, but I'm not sure what made it in, so I'm just guessing how it works.

"Allows the QP planner to re-plan a trajectory if Astrobee encounters an obstacle." -> "Allows Astrobee to replan if it detects an obstacle too close to its forward trajectory. Enabling replanning only makes sense when you are using a planner that is able to plan around obstacles. (See Settings.setPlanner. As of 5/2021, only the QP planner can plan around obstacles)."

"Set to true/false to enable/disable re-planning." -> "If true, when Astrobee detects an obstacle too close to its forward trajectory, after the robot comes to a stop, the choreographer will automatically request a new trajectory from the currently configured planner. If false, the robot will stop and wait for operator assistance. "

kbrowne15:
Thanks for the comment! The descriptions have been changed.

PR #183, opened 2021-05-24
Rotation fallback fix 

rsoussan:
Includes refactoring that cleans up code and allows for potential split graph architecture for localization and VIO.  The rotation fix enables a fallback when visual odometry factors are invalid that creates a rotation only factor.


localization/graph_localizer/src/feature_point_graph_values.cc
```diff
@@ -0,0 +1,121 @@
...
+    if (num_factors <= 0) {
+      old_features.emplace_back(key);
+    }
+  }
+  return old_features;
+}
+
+void FeaturePointGraphValues::RemoveOldFeatures(const gtsam::KeyVector& old_keys) {
+  for (const auto& key : old_keys) {
+    // TODO(rsoussan): test this
```
bcoltin, line 108: Did we test this?

bcoltin (approved):
It looks like there are a lot of TODO's, can you double check them? 

Then if you fix the lint error and build error I think we are good.

marinagmoreira:
Can you fix the lint errors and:

CMakeFiles/graph_optimizer.dir/src/graph_optimizer.cc.o: In function `graph_optimizer::GraphOptimizer::OldKeysAndFactors(double)':
/src/astrobee/localization/graph_optimizer/src/graph_optimizer.cc:125: undefined reference to `graph_optimizer::RemoveOldFactors(std::vector<unsigned long, std::allocator<unsigned long> > const&, gtsam::NonlinearFactorGraph&)'
collect2: error: ld returned 1 exit status
localization/graph_optimizer/CMakeFiles/graph_optimizer.dir/build.make:359: recipe for target 'devel/lib/libgraph_optimizer.so' failed

localization/graph_localizer/include/graph_localizer/robust_smart_projection_pose_factor.h
```diff
@@ -75,34 +75,26 @@ class RobustSmartProjectionPoseFactor : public SmartProjectionPoseFactor<CALIBRA
...
   }
 
   double error(const Values& values) const override {
     if (this->active(values)) {
       try {
         const double total_reprojection_loss = this->totalReprojectionError(this->cameras(values));
         const auto result = this->point();
-        // TODO(rsoussan): This should be here but leads to a degeneracy (see comment in linearize)
-        // if (!result.valid() && !useForRotationOnly(result)) return 0.0;
-        // Multiply by 2 since totalReporjectionError divides mahal distance by 2, and robust_model_->loss
```
marinagmoreira, line 104: Did you want to remove the // Multiply by 2 (...) explanation, or deleted by accident?

localization/graph_localizer/src/feature_point_graph_values.cc
rsoussan: This was deprioritized because its for sliding window bundle adjustment, which is pretty experimental right now

localization/graph_localizer/include/graph_localizer/robust_smart_projection_pose_factor.h
rsoussan: Good catch, I've added this back

rsoussan:
Fixed the lint error, I forgot to add a src file

marinagmoreira (approved):

PR #184, opened 2021-05-25
Update platform to new version.

bcoltin:


bcoltin:
Went ahead and merged this myself since it is breaking things internally / won't affect external people / is very simple.

PR #187, opened 2021-05-26
Added enable re-plan to mobility settings telemetry.

kbrowne15:
Added what enable re-plan is set to in the executive to the mobility settings telemetry message. I should have done this when I added the enable re-plan command but forgot.

### DIFF
--- a/communications/dds_msgs/idl/MobilitySettingsState.idl
+++ b/communications/dds_msgs/idl/MobilitySettingsState.idl
@@ -66,6 +66,11 @@ module rapid {
 
         //@copy-declaration /** Sets the planner mobility uses. */
         public string planner;
+
+        //@copy-declaration /** Setting this to true allows Astrobee to re-plan
+        //@copy-declaration   * if there is an obstacle to close to the robot's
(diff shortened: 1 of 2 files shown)


bcoltin (approved):

bcoltin:
@kbrowne15 feel free to merge once the build passes

trey0:
Do any subscribers need to be modified, or at least recompiled, to maintain compatibility after this change? (Like does the Facility team need to update ground tools?)

kbrowne15:
Yes, the IDLs need to be recompiled for GDS and it is on my list of things to do. However the mobility settings message was never "manually" added to GDS so it won't break anything in GDS and only shows up in the debug tab in the raw telemetry window. Since the facility doesn't use this window, it is not high priority.

trey0:
Thanks, that makes sense. Is it worth double-checking that publishing messages with the new IDL doesn't surprise-crash a GDS control station compiled with the old IDL, even though we think GDS "shouldn't be using it" outside of rare corner cases?

kbrowne15:
I already tested it before the pr and it doesn't crash gds.

PR #198, opened 2021-06-07
Add compatibility with ubuntu 20

marinagmoreira:
- Change the version names from the ROS version to the ubuntu version, seems to make more sense
- It compiles now with C++14 as it is the minimum for ubuntu 20
- Compatible with ubuntu 20


external/gtest/CMakeLists.txt
```diff
@@ -28,8 +28,15 @@ CMakeExternalProject(
...
-    ${gtest_noros_BINARY_DIR}/libgtest_main${CMAKE_SHARED_LIBRARY_SUFFIX})
+
+  if("${ROS_DISTRO}" STREQUAL "noetic")
+    set(GTEST_LIBRARIES ${gtest_noros_BINARY_DIR}/lib/libgtest${CMAKE_SHARED_LIBRARY_SUFFIX}
+  ${gtest_noros_BINARY_DIR}/lib/libgtest_main${CMAKE_SHARED_LIBRARY_SUFFIX})
+  else()
+    set(GTEST_LIBRARIES ${gtest_noros_BINARY_DIR}/libgtest${CMAKE_SHARED_LIBRARY_SUFFIX}
+  ${gtest_noros_BINARY_DIR}/libgtest_main${CMAKE_SHARED_LIBRARY_SUFFIX})
+  endif()
+
```
marinagmoreira, line 39: This is very hacky, but we went around for far too long and couldn't get a better alternative

scripts/docker/astrobee_base_bionic.Dockerfile
```diff
@@ -43,7 +43,8 @@ COPY ./scripts/setup/debians /setup/astrobee/debians
 
 RUN apt-get update \
   && /setup/astrobee/debians/build_install_debians.sh \
-  && rm -rf /var/lib/apt/lists/*
+  && rm -rf /var/lib/apt/lists/* \
+  && rm -rf /setup/astrobee/debians
```
marinagmoreira, line 47: This actually reduced the size of the image by half!
I was getting insufficient space on device on the github CI run -.-

CMakeLists.txt
```diff
@@ -114,12 +114,16 @@ if (USE_CTC)
     list(APPEND CMAKE_PREFIX_PATH "${ARM_CHROOT_DIR}/opt/ros/kinetic")
   elseif("${ROS_DISTRO}" STREQUAL "melodic")
     list(APPEND CMAKE_PREFIX_PATH "${ARM_CHROOT_DIR}/opt/ros/melodic")
+  elseif("${ROS_DISTRO}" STREQUAL "noetic")
```
bcoltin, line 117: Can I suggest: list(APPEND CMAKE_PREFIX_PATH "${ARM_CHROOT_DIR}/opt/ros/${ROS_DISTRO}")

CMakeLists.txt
```diff
@@ -114,12 +114,16 @@ if (USE_CTC)
...
+    list(APPEND CMAKE_PREFIX_PATH "${ARM_CHROOT_DIR}/opt/ros/noetic")
   endif()
 else (USE_CTC)
   if("${ROS_DISTRO}" STREQUAL "kinetic")
     list(APPEND CMAKE_PREFIX_PATH "/opt/ros/kinetic")
   elseif("${ROS_DISTRO}" STREQUAL "melodic")
     list(APPEND CMAKE_PREFIX_PATH "/opt/ros/melodic")
+  elseif("${ROS_DISTRO}" STREQUAL "noetic")
```
bcoltin, line 125: same as last comment

CMakeLists.txt
```diff
@@ -134,6 +138,8 @@ if("${ROS_DISTRO}" STREQUAL "kinetic")
   list(APPEND CMAKE_INSTALL_RPATH "/opt/ros/kinetic/lib")
 elseif("${ROS_DISTRO}" STREQUAL "melodic")
   list(APPEND CMAKE_INSTALL_RPATH "/opt/ros/melodic/lib")
+elseif("${ROS_DISTRO}" STREQUAL "noetic")
```
bcoltin, line 141: again

CMakeLists.txt
```diff
@@ -342,6 +348,8 @@ if (USE_ROS)
       set(OROCOS_KDL_LIBRARIES ${ARM_CHROOT_DIR}/opt/ros/kinetic/lib/liborocos-kdl.so)
     elseif("${ROS_DISTRO}" STREQUAL "melodic")
       set(OROCOS_KDL_LIBRARIES ${ARM_CHROOT_DIR}/opt/ros/melodic/lib/liborocos-kdl.so)
+    elseif("${ROS_DISTRO}" STREQUAL "noetic")
```
bcoltin, line 351: actually, maybe for all of these you can set a $ROS_DIR to simplify things greatly :)

localization/lk_optical_flow/src/lk_optical_flow.cc
```diff
@@ -25,6 +25,8 @@
 #include <opencv2/video/tracking.hpp>
 #include <ros/ros.h>
 
+#define CV_FONT_NORMAL CV_RAND_NORMAL
```
bcoltin, line 28: I don't think this is what you want to do... it probably compiles since they are both ints but I think they are unrelated things, one is a font weight and one is a normal distribution?

localization/marker_tracking/include/marker_tracking/arxmlio.h
```diff
@@ -35,7 +36,7 @@ namespace marker_tracking {
   // 3 - TL
   typedef std::map<int, Eigen::Matrix<float, 4, 3>,
                    std::less<int>,
-                   Eigen::aligned_allocator<std::pair<int, Eigen::Vector4f> > >
+                   std::allocator<std::pair<int const, Eigen::Matrix<float, 4, 3>> > >
```
bcoltin, line 39: what happened here? have to be careful https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html, this bug might only show up on arm sometimes if I recall

localization/marker_tracking/include/marker_tracking/marker_detector.h
```diff
@@ -19,6 +19,12 @@
 #ifndef MARKER_TRACKING_MARKER_DETECTOR_H_
 #define MARKER_TRACKING_MARKER_DETECTOR_H_
 
+#ifndef CV__ENABLE_C_API_CTORS
+#define CV__ENABLE_C_API_CTORS
+#endif
+#include <opencv2/imgproc.hpp>
```
bcoltin, line 25: usually better not to put this in the header unless we really need to

localization/sparse_mapping/include/sparse_mapping/tensor.h
```diff
@@ -47,7 +48,7 @@ namespace cv {
 namespace sparse_mapping {
 
   typedef std::map<std::pair<int, int>, Eigen::Affine3d, std::less<std::pair<int, int> >,
-                   Eigen::aligned_allocator<std::pair<std::pair<const int, const int>, Eigen::Affine3d> > >
+                   std::allocator<std::pair<std::pair<int , int > const, Eigen::Affine3d> > >
```
bcoltin, line 51: could cause problems

scripts/setup/debians/build_install_debians.sh
```diff
@@ -43,26 +43,55 @@ elif [ "$DIST" = "bionic" ]; then
...
+
+  # opencv
+  cd ${DEBIAN_LOC}/opencv
+  sudo mk-build-deps -i -r -t "apt-get --no-install-recommends -y" control
+  cd ${DEBIAN_LOC}
+  ./build_opencv.sh || exit 1
+  cd ${DEBIAN_LOC}/libopencv
+  sudo dpkg -i libopencv*_amd64.deb || exit 1
+
+  # alvar
```
bcoltin, line 76: it seems like there is a lot of overlap between these two ubuntu versions... can we make them share more of the script?

bcoltin (changes requested):

localization/lk_optical_flow/src/lk_optical_flow.cc
marinagmoreira: Oh wow, I got tricked by the:
/home/mgouveia/ros_ws/astrobee/localization/lk_optical_flow/src/lk_optical_flow.cc:221:17: error: ‘CV_FONT_NORMAL’ was not declared in this scope; did you mean ‘CV_RAND_NORMAL’?

localization/marker_tracking/include/marker_tracking/marker_detector.h
marinagmoreira: That function is in opencv2/imgproc.hpp, but it will only be defined if CV__ENABLE_C_API_CTORS is enabled

bcoltin: It doesn't seem like RGB is called in the header though? Is there a reason we can't put it in the C++ function?

marinagmoreira: Another solution would be to modify the alvar repo instead

bcoltin: Huh, I wonder why alvar compiles then... I guess this is fine, maybe make a comment why / when this is needed.

localization/marker_tracking/include/marker_tracking/arxmlio.h
marinagmoreira: Thanks for the template, I followed it and didn't get compile errors (was following the std one before. I'm not sure how this bit worked before as it was trying to put a vector 4f into a matrix 4x3...

bcoltin (approved):
Good to merge once it passes tests!

KhaledSharif:
awesome work!!

PR #202, opened 2021-06-10
add fix for expired ROS apt key. 

trusktr:
Info: https://discourse.ros.org/t/ros-gpg-key-expiration-incident/20669

I'm not sure if this was fixed already (I didn't see something in the commits, unless I missed it), but this is what I had to do to get things working again. I'm also not sure if this is the best way to fix it.

cc @KhaledSharif @marinagmoreira @bcoltin

### DIFF
--- a/scripts/setup/add_ros_repository.sh
+++ b/scripts/setup/add_ros_repository.sh
@@ -24,3 +24,5 @@ sudo apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31
 # Gazebo stuff
 sudo sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable $(lsb_release -sc) main" > /etc/apt/sources.list.d/gazebo-stable.list'
 wget http://packages.osrfoundation.org/gazebo.key -O - | sudo apt-key add
+
+curl -s https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | sudo apt-key add -
--- a/scripts/docker/astrobee_base_kinetic.Dockerfile
+++ b/scripts/docker/astrobee_base_kinetic.Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get update \
     lsb-release \
     sudo \
     wget \
+    curl \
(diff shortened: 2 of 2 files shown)


scripts/setup/add_ros_repository.sh
```diff
@@ -24,3 +24,5 @@ sudo apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31
 # Gazebo stuff
```
bcoltin, line 24: Please use ubuntu keyserver like beofre with new key id (more trustworthy than downloading from random site) and replace the old key. Also no need to add curl to dependencies.

trusktr: Indeed I added `curl` to the deps list.

I'm not sure what the apt-key command should look like. I see that this page, http://wiki.ros.org/Installation/Ubuntu/Sources, appears to have the outdated key.

trusktr: Oh, I misread, you said no need to add curl. I thought it failed, but maybe I mis-recall. Let me try...

trusktr: I tried again, and `curl` needed to be installed for the `curl` command to work.

That http://wiki.ros.org/Installation/Ubuntu/Sources page does seem to be outdated, and I am waiting for them to reply with an update.

bcoltin: I am saying to use wget which is already installed and used in the previous line instead of curl. They do the same thing.

You also need to remove the old key installation - C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654

trusktr: Alright, the ROS team responded, and they said that the original command should still work. I think the issue was I updated the `Dockerfile` at a point _after_ that `apt-key` command, so its result remained cached. `docker build --no-cache` is one way to build fresh again.

Trying a rebuild from scratch with the original command now.

bcoltin:
@trusktr is the original command working again?

trusktr:
@bcoltin Yeah, worked after I blew away docker cache and worked out the other issues.

PR #204, opened 2021-06-11
Simulation PMC Timeout Update

Pedro-Roque:
Added support for changing the PMC timeout on the simulation, to support changing the cutoff frequency of the PMC in Gazebo too. 

@bcoltin


bcoltin (approved):
Wow, we can't spell... Thanks!

PR #206, opened 2021-06-14
Cargo demo

marinagmoreira:
Changes for the cargo demo to work with develop branch


astrobee/config/management/executive.config
```diff
@@ -24,7 +24,7 @@ led_service_available_timeout = 10
 
 gs_command_timeout = 4
 
-motion_feedback_timeout = 1
+motion_feedback_timeout = 5
```
marinagmoreira, line 27: replanning takes longer than 1 second sometimes because of the iterations

astrobee/config/mobility/planner_qp.config
```diff
@@ -39,10 +39,10 @@ parameters = {
     id = "robot_radius",
     reconfigurable = true,
     type = "double",
-    default = 0.16,
+    default = 0.22,
     min = 0.0,
     max = 1.0,
-    description = "Radius of the robot considered in the planner (default is 0.15 * 2* sqrt(3))",
+    description = "Radius of the robot considered in the planner (default is 0.15 * sqrt(2))",
```
marinagmoreira, line 45: The previous parameter was not enough, I changed it to 0.15*sqrt(2) to account for the diagonal

mobility/choreographer/src/choreographer_nodelet.cc
```diff
@@ -291,6 +291,7 @@ class ChoreographerNodelet : public ff_util::FreeFlyerNodelet {
         }
         states_.insert(states_.begin(), current_pose);
         replan_attempts_++;
+        ros::Duration(1).sleep();  // sleep for a second
```
marinagmoreira, line 294: This is necessary in between planning attempts in case the obstacle map has some old obstacles that are not valid anymore, it allows time to update the map

astrobee/config/mobility/planner_qp.config
trey0: As we just discussed in chat, to find the radius of the circumscribing sphere of a cube with width ~0.32, the formula would be 0.16 * sqrt(3) which comes to 0.28 (rounding up).

astrobee/config/management/executive.config
bcoltin: what do you mean by the iterations?

marinagmoreira: replanning has a 1s delay for the octomap to be updated and forget the non-visible obstacles

mobility/choreographer/src/choreographer_nodelet.cc
bcoltin: This goes against the spirit of the finite state machine, which should have very quick functions in this part... with this sleep, the state machine can't be interrupted for example with a cancel event.

If we are going to do this, we should have a separate sleeping state and have an event trigger the FSM when a timer goes off.

However, I think this is a very special case where we expect the cargo to move. I kind of think instead of putting this wait in the choreographer it might make sense to try multiple times at the behavior level. In the general case, when we don't expect things to move, it would be better to fail quickly. Or we could make it so we can change the wait time and retries when we call the choreographer--- maybe that is the best option, but I'm not sure what changes it would take. What do you think?

marinagmoreira: This is a problem that actually shows up in all qp planner trajectories and not specific to the cargo demo. When the trajectory detects obstacles and fails, sometimes when the obstacle map is updated because of the "memory_time" function, the objects previously detected are updated and the robot is able to continue instead of getting stuck. This delay allows the mapper to get rid of older obstacles that are not in the fov of the robot.

Given this, we should probably go with approach 1?

bcoltin (approved):

PR #209, opened 2021-06-16
Add python3 flag

marinagmoreira:
Turns out the tf package in kinetic can only be used by python2.
I'm not sure if there is a way of automatically checking the ros version on launchfile to avoid all these arguments.

Related issues:
https://github.com/RobotnikAutomation/summit_xl_sim/issues/21
https://answers.ros.org/question/326226/importerror-dynamic-module-does-not-define-module-export-function-pyinit__tf2/


trey0:
Without fully understanding everything you're doing here, I noticed you created split scripts `spawn_model` vs. `spawn_model_python3`, which could be a pain from a maintenance perspective. Given the simple differences I see between them, I think you could easily create a unified script that works under both Python 2 and 3 using this library: https://six.readthedocs.io/

trey0:
> Without fully understanding everything you're doing here, I noticed you created split scripts `spawn_model` vs. `spawn_model_python3`, which could be a pain from a maintenance perspective. Given the simple differences I see between them, I think you could easily create a unified script that works under both Python 2 and 3 using this library: https://six.readthedocs.io/

In fact, the differences are so minor that you may be able to make a unified script without introducing a dependency on the six library. See [1] [2]

[1] https://www.semicolonworld.com/question/59028/how-to-use-from-future-import-print-function
[2] https://stackoverflow.com/questions/11285313/try-except-as-error-in-python-2-5-python-3-x

PR #217, opened 2021-06-21
Handrail rebase with lint and compile fixes

rsoussan:
Same as PR 215 (https://github.com/nasa/astrobee/pull/215) but with lint and compile fixes as I was not able to push directly to that PR.


bcoltin (approved):
Good to merge once you pass the continuous integration

PR #219, opened 2021-06-25
Fixed loc rviz plugins

rsoussan:
Building of loc rviz plugins was disabled after the localizer code was refactored.  It has now been updated with the latest localization code and works again.


bcoltin (approved):
Fix build and then it looks good.

PR #222, opened 2021-06-29
Add load and unload nodelet command to the executive.

kbrowne15:
Also fixed some node names and types in the system monitor config.


management/executive/src/executive.cc
```diff
@@ -3348,6 +3434,82 @@ bool Executive::Undock(ff_msgs::CommandStampedPtr const& cmd) {
   return docked;
 }
 
+bool Executive::UnloadNodelet(ff_msgs::CommandStampedPtr const& cmd) {
+  NODELET_INFO("Executive executing unload nodelet command!");
+  bool success = true;
+  std::string err_msg = "Unload nodelet parameters must be strings.";
+  ff_msgs::UnloadLoadNodelet unload_load_nodelet_srv;
+  unload_load_nodelet_srv.request.load = false;
```
bcoltin, line 3442: are these two functions just the same with this set to true or false? if so should combine them with a boolean argument

management/executive/src/executive.cc
```diff
@@ -3348,6 +3434,82 @@ bool Executive::Undock(ff_msgs::CommandStampedPtr const& cmd) {
...
+
+  // Don't unload a nodelet while moving because we may be unloading a mobility
+  // nodelet
+  if (CheckNotMoving(cmd)) {
```
bcoltin, line 3446: What happens if we are moving? It doesn't look like it gives an error with an error message?

bcoltin (changes requested):

management/executive/src/executive.cc
kbrowne15: The check not moving function sends a failed ack i.e. it does give an error message.

management/executive/src/executive.cc
kbrowne15: Good point!

management/executive/src/executive.cc
bcoltin: Is that the right behavior? Seems like it could be confusing.

kbrowne15: This function is only used to check if the robot is not moving in terms of executing a command. It seemed cleaner to ack in the function rather than having the same ack if chain in multiple command functions. Would it be clearer if the name was changed to CheckNotMovingAndFailAckIfMoving?

management/executive/src/executive.cc
kbrowne15: I combined the code into one function and tried to make it cleaner.

management/executive/src/executive.cc
kbrowne15: Renamed CheckNotMoving to FailCommandIfMoving.

bcoltin (approved):

PR #223, opened 2021-07-06
Fixed not removing faulty factors bug

rsoussan:
Small fix to correct corner case where invalid factors are still being added to the graph.


bcoltin (approved):

bcoltin:
Can merge once it passes tests

marinagmoreira (approved):

PR #226, opened 2021-07-10
Optionally publish NavCam and DockCam images in raw Bayer format that can later be de-Bayered to RGB

trey0:
Not tested yet!

Adds a feature to the is_camera driver for NavCam and DockCam that optionally publishes images in raw Bayer format. The feature is disabled by default in cameras.config (bayer_enable parameter). The rate of publishing Bayer images can be throttled (bayer_throttle_ratio parameter).

The implementation is much like the existing grayscale output but instead of converting Bayer to grayscale, it just copies the raw Bayer imagery into the output message (ROS supports tagging the image as Bayer encoded).

There are a number of ways we could use it downstream. Perhaps the most ROS-idiomatic approach would be to run the Bayer-encoded images through the ROS image_proc package's debayer nodelet, and then subscribe to the resulting color image topic in the image sampler.


trey0:
Oops, PR should have been to develop.

PR #228, opened 2021-07-12
fixing carriage tf prefix

marinagmoreira:
I was testing the granite sim and noticed I missed this bit while adding the prefix manually to the urdf for ubuntu 20 compatibility


bcoltin (approved):

bcoltin:
Can merge once tests pass

PR #249, opened 2021-08-02
convert python 2to3, adding python linter to CI

marinagmoreira:
- Made all python scripts go through 2to3 tool, to make it ready to run in python3 (mostly all print changes)
- Adding python linter check to CI (black and isort)
- Added python linter check to pre-commit file


scripts/git/pre-commit.linter_python
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+files=$(git diff --cached --name-only | grep '\.py$')
+
+# If files is empty exit success
+if [ -z "${files}" ]; then
+	exit 0
```
trey0, line 24: echo "No Python files changed, no checks needed."

trey0 (commented):
Thanks, looks great!

bcoltin:
I want to get @rsoussan's opinion on this since he writes the most python code. It seems like you were using two spaces Ryan but the linter is requiring four?

Marina, do you feel strongly about using four?

trey0:
> I want to get @rsoussan's opinion on this since he writes the most python code. It seems like you were using two spaces Ryan but the linter is requiring four?

Python has a standard style guide, PEP 8 [1], which gives guidelines on some but not all style questions. It does call for four space indentation.

Also, for context, I think Marina is using the Black Python auto-formatting tool [2] on my recommendation. Black has fairly strong opinions about how to format things and is not designed to allow much customizability. It mostly matches PEP 8 but also has opinions on issues where PEP 8 is silent. The lack of customizability can be annoying in some ways, but is also nice in that we can't be tempted to fiddle with it too much, and using Black will tend to make our style exactly match other Black users.

[1] https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces
[2] https://black.readthedocs.io/en/stable/

marinagmoreira:
> I want to get @rsoussan's opinion on this since he writes the most python code. It seems like you were using two spaces Ryan but the linter is requiring four?
> 
> Marina, do you feel strongly about using four?

I don't have any preferences as I don't use python much

rsoussan:
I also don't have a preference, thanks for asking though

rsoussan (approved):

bcoltin (approved):

PR #257, opened 2021-08-06
Uniform collision check

marinagmoreira:
- Made services in mapper more explicit and added services to check the variables that are defined through services
- Inflation now defined as the robot radius + collision distance
- Connected collision distance from the executive in a plan to the mapper
- Zones map in validator now use voxel map, allowing inflation
- QP planner uses validator map instead of recalculating zones (added a service to get the zone map if needed by other planners). 'planner_distance' is now the extra inflation distance defined only for the planner, allowing for extra planning distance from obstacle to obtain trajectories further from walls.


mobility/choreographer/readme.md
```diff
@@ -59,4 +60,57 @@ To ensure that the robot is following the desired trajectory, the choreographer
 | ```tolerance_vel```              | Velocity tolerance checked thoughout the entire trajectory |
 | ```tolerance_att```              | Attitude tolerance checked thoughout the entire trajectory |
 | ```tolerance_omega```            | Omega tolerance checked thoughout the entire trajectory. |
-| ```tolerance_time```             | Time for sync check |
\ No newline at end of file
+| ```tolerance_time```             | Time for sync check |
+
+The planned trajectories are also checked against the keep-in and keep-out zones.
```
bcoltin, line 65: nice documentation!

mobility/choreographer/src/validator.cc
```diff
@@ -22,17 +22,150 @@
...
+bool Validator::GetZonesMap() {
+    ff_msgs::GetFloat srv;
+    if (get_resolution_.call(srv)) {
+      map_res_ = srv.response.data;
+    }
+
+
+    // Check min and max of keepin zones to use as map boundries
+    Vec3f min, max, zmin, zmax;
+    min << 1000.0, 1000.0, 1000.0;
```
bcoltin, line 35: Can use std::numeric_limits max/min or infinity

mobility/choreographer/src/validator.cc
```diff
@@ -22,17 +22,150 @@
...
+    int num_cell = dim(0) * dim(1) * dim(2);
+
+
+    // Declare voxel map
+    std::vector<signed char> map(num_cell, val_unknown_);
+    jps_map_util_.reset(new JPS::VoxelMapUtil());
+    jps_map_util_->setMap(origin, dim, map, map_res_);
+
+    // Keepin/Keepout zones:
+    // To reduce computational load, only the contourn of the keepin/keepout
```
bcoltin, line 77: contourn?

mobility/choreographer/src/validator.cc
```diff
@@ -22,17 +22,150 @@
...
+    // case where keepin zones are connected and a contourn was put between them
+    for (auto &zone : zones_.zones) {
+      if (zone.type == ff_msgs::Zone::KEEPIN) {
+        zmin << std::min(zone.min.x, zone.max.x),
+            std::min(zone.min.y, zone.max.y), std::min(zone.min.z, zone.max.z);
+        zmax << std::max(zone.min.x, zone.max.x),
+            std::max(zone.min.y, zone.max.y), std::max(zone.min.z, zone.max.z);
+        // add points on surface
+        Vec3f tmp = Vec3f::Zero();
+        for (auto zx = zmin(0); zx <= zmax(0); zx += map_res_) {
```
bcoltin, line 117: if it's fast enough it doesn't matter, but since there are probably fewer keepin zones than points in the keepin zones, might be faster to check the points as you add them instead

mobility/choreographer/src/validator.cc
```diff
@@ -22,17 +22,150 @@
...
+              tmp(i) = zmax(i);
+              map[jps_map_util_->getIndex(jps_map_util_->floatToInt(tmp))] =
+                  val_occ_;
+            }
+          }
+        }
+      }
+    }
+
+    // set voxel map using keepin/keepout zones
```
bcoltin, line 159: If you can, might want to split this into multiple functions for the different steps

mobility/choreographer/src/validator.cc
```diff
@@ -22,17 +22,150 @@
...
+    jps_map_util_->setMap(origin, dim, map, map_res_);
+
+    // 6) Inflate using the robot radius and the collision distance
+    double inflation = map_res_;
+    if (get_resolution_.call(srv)) {
```
bcoltin, line 164: why are you getting the map resolution again and using it to inflate? or am I missing something

bcoltin: guessing you meant to use get_map_inflation_ ?

mobility/planner_qp/readme.md
```diff
@@ -43,7 +43,7 @@ The planner has adjustable parameters which can affect the optimizer's performan
 
 * `sample_rate` - The optimization uses a continuous trajectory representation which is sub-sampled at this rate before being sent to the choreographer.
 
-* `robot_radius` - How large the robot's center must stay away from obstacles, increase this to create a trajectory further away from walls
+* `planner_distance` - Additional distance that the planner should stay away from occupied areas. The occupied areas are defined by the zones and obstacles inflated using the robot radius plus the collision distance. The minimum planner distance is set to the map resolution, a higher planner distance implies that the trajectory generated will pass further away from walls.
```
bcoltin, line 46: Maybe explain briefly why we needed this

bcoltin (changes requested):
Looks great! All very minor suggestions (up to you) with the exception of one potential bug.

bcoltin:
@kbrowne15 if you could please still take a look at the executive part!

communications/ff_msgs/srv/GetZonesMap.srv
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2017, United States Government, as represented by the
```
kbrowne15, line 1: This looks good but I'm wondering if we should rename the service. We already have a get zones service and I worry that it isn't clear what the difference is between the two. What do you think about GetOccupancyMap?

kbrowne15 (approved):
Looks good once you consider my minor comment and Brian's suggestions.

communications/ff_msgs/srv/GetZonesMap.srv
marinagmoreira: I think that's a good idea. Initially I was skeptical because the 'occupancy map' name might inadvertently infer that we also have the obstacle map contents in it, but the service name should be the one clarifying the contents of the message and not the message name.

mobility/choreographer/src/validator.cc
marinagmoreira: I'm not sure I understand, we only iterate through the points belonging to the zone

mobility/choreographer/src/validator.cc
marinagmoreira: thank you for spotting that :)

mobility/choreographer/src/validator.cc
bcoltin: My understanding is, right now you have:

1) Set all to unoccupied
2) Set borders to occupied
3) Set keepins to unoccupied
4) Set keepouts to occupied

I suspect it might be faster to instead do:

1) Set all to unoccupied
2) Set borders to occupied, while looping through keepin zones to check they are not in one
3) Set keepouts to occupied

Since the number of pixels inside all zones is probably bigger than the number of border pixels times the number of zones.

But if it's already fast enough no need to worry about it.

marinagmoreira: But in the second approach we would still have to fill in the keepin zones as free

bcoltin: Oh, I thought the map started with everything set to free.

marinagmoreira: I made it start as val_unknown_, so we are able to throw the error VIOLATES_KEEP_IN

bcoltin: Ok sounds good!

bcoltin (approved):

PR #259, opened 2021-08-12
End of motion check

marinagmoreira:


bcoltin (commented):
Do you think setpoint is the best term? I'm not sure if I would know what that means... What about endpoint?

trey0:
> Do you think setpoint is the best term? I'm not sure if I would know what that means... What about endpoint?

I agree. I think "setpoint" is pretty common in the controls community but not familiar to everyone, and "endpoint" is more specific anyway.

PR #260, opened 2021-08-12
explicitly specify python3 while running for ubuntu 16 machines

marinagmoreira:
While running the hook on my ubuntu 16 machine I got an import error, seems like the library we had to upgrade because of python3/ubuntu 20 requirements is not back compatible. Therefore I made it use python3 instead of the default.

### DIFF
--- a/scripts/git/pre-commit.linter_cpp
+++ b/scripts/git/pre-commit.linter_cpp
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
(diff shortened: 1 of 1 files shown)


bcoltin:
Is python3 installed by default on 16.04? If so looks good!

marinagmoreira:
I have a 16.04 vm that the only thing I ever installed was astrobee and has python3, so I think it does :)

bcoltin (approved):

PR #265, opened 2021-08-19
Add clock_sync_monitor.sh

trey0:


bcoltin:
Might make more sense in astrobee_ops repo, what do you think @kbrowne15 ?

trey0:
> Might make more sense in astrobee_ops repo, what do you think @kbrowne15 ?

That was my first thought as well, but when I looked into astrobee_ops I saw the scripts are organized by what processor they run on, and as far as I can tell none of them run onboard the robot.

Another option would be to put the script in this repo in the scripts/daemon/astrobee folder, although the style of this script is quite different from the other scripts in that folder.

If there is a convenient/standardized way to install this script on the mlp, that would actually be the preferred place to run it from (changing the list of remote hosts it checks to be the hlp and llp.)

PR #267, opened 2021-08-24
update dockerfile UBUNTU_VERSION values...

trusktr:
...so that the version number can be used downstream in isaac dockerfiles without the 'ubuntu' prefix. Also update README regarding the dockerfiles.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -20,14 +20,14 @@ jobs:
 
     - name: Build code for Ubuntu 16
       run: docker build . -f ./scripts/docker/astrobee.Dockerfile
-                          --build-arg UBUNTU_VERSION=ubuntu16.04
+                          --build-arg UBUNTU_VERSION=16.04
                           --build-arg ROS_VERSION=kinetic
                           --build-arg PYTHON=''
                           -t astrobee/astrobee:latest-ubuntu16.04
 
     - name: Test code
       run: docker build . -f ./scripts/docker/test_astrobee.Dockerfile
-                          --build-arg UBUNTU_VERSION=ubuntu16.04
+                          --build-arg UBUNTU_VERSION=16.04
                           -t astrobee/astrobee:test-ubuntu16.04
 
   build-bionic:
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -42,14 +42,14 @@ jobs:
 
     - name: Build code for Ubuntu 18
       run: docker build . -f ./scripts/docker/astrobee.Dockerfile
-                          --build-arg UBUNTU_VERSION=ubuntu18.04
+                          --build-arg UBUNTU_VERSION=18.04
                           --build-arg ROS_VERSION=melodic
                           --build-arg PYTHON=3
                           -t astrobee/astrobee:latest-ubuntu18.04
 
     - name: Test code
       run: docker build . -f ./scripts/docker/test_astrobee.Dockerfile
-                          --build-arg UBUNTU_VERSION=ubuntu18.04
+                          --build-arg UBUNTU_VERSION=18.04
                           -t astrobee/astrobee:test-ubuntu18.04
 
   build-focal:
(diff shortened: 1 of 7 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -20,14 +20,14 @@ jobs:
...
-                          --build-arg UBUNTU_VERSION=ubuntu16.04
+                          --build-arg UBUNTU_VERSION=16.04
                           --build-arg ROS_VERSION=kinetic
                           --build-arg PYTHON=''
                           -t astrobee/astrobee:latest-ubuntu16.04
 
     - name: Test code
       run: docker build . -f ./scripts/docker/test_astrobee.Dockerfile
-                          --build-arg UBUNTU_VERSION=ubuntu16.04
+                          --build-arg UBUNTU_VERSION=16.04
```
marinagmoreira, line 30: Needs to change the definitions for ubuntu 18 and 20 below too

scripts/docker/readme.md
```diff
@@ -12,13 +12,18 @@ The docker image for the astrobee FSW is divided throughout 2 docker files.
...
-    astrobee_base_focal - Contains installation of all astrobee dependencies the Ubuntu 20.04 + ROS noetic setup.
+- `astrobee_base.Dockerfile` - Contains installation of all astrobee dependencies the Ubuntu + ROS setup.
+- `astrobee.Dockerfile` - Builds the astrobee FSW code on top of astrobee_base.
 
-    astrobee_xenial - Builds the astrobee FSW code on top of astrobee_base_xenial.
-    astrobee_bionic - Builds the astrobee FSW code on top of astrobee_base_bionic.
-    astrobee_focal - Builds the astrobee FSW code on top of astrobee_base_focal.
+The Docker files accept the following version args (note that they must match up):
+
+- `UBUNTU_VERSION` - The version of Ubuntu to use. Valid values are "16.04", "18.04", and "20.04". Defautls to "16.04".
```
marinagmoreira, line 20: It actually defaults to the current OS in the build.sh with:
os=`cat /etc/os-release | grep -oP "(?<=VERSION_CODENAME=).*"`

marinagmoreira (changes requested):

scripts/docker/readme.md
trusktr: Oh yeah. I was thinking of the `ARG`s inside the docker files that default to `16.04`. Should add a blurb about `build.sh`. My system is `manjaro` so I gotta specify the arg for `build.sh` in my case

.github/workflows/build_and_test.yml
trusktr: Thanks, fixed.

scripts/docker/readme.md
trusktr: Oh, I forgot: because my OS is Manjaro Linux, then it still falls back to the default of `16.04`/`kinetic` because the conditional checks only try to match against bionic or focal. So the script will use xenial if the OS is not bionic or focal, which works out. In my case the default is xenail.

marinagmoreira: Astrobee doesn't currently support any other operating system besides ubuntu. (https://nasa.github.io/astrobee/html/install-nonNASA.html)

trusktr: True that! That's why I'm using the Docker containers. Everything is running in Ubuntu inside the Docker containers.

I was just mentioning that because I'm not on Ubuntu when I run `build.sh`, the [conditional check in `build.sh`](https://github.com/nasa/astrobee/pull/267/files#diff-be45b1c53fa8ef7864c013a77325ed29aca3eeae2be39a37e107d63e0d6134d1L55-L68) will default to `xenial` for the docker container because of the [`os` variable creation](https://github.com/nasa/astrobee/pull/267/files#diff-be45b1c53fa8ef7864c013a77325ed29aca3eeae2be39a37e107d63e0d6134d1R31).

trusktr: Hmmm.... But maybe I should just make a generic Ubuntu container to run all the containers in. That will probably work best. I haven't tried that yet, I've just been running all these containers directly in Manjaro Linux.

marinagmoreira: I think your setup is fine, you just need to be aware that it is not the standard and the documentation will not reflect this use case.

scripts/docker/readme.md
```diff
@@ -12,28 +12,36 @@ The docker image for the astrobee FSW is divided throughout 2 docker files.
...
-    astrobee_focal - Builds the astrobee FSW code on top of astrobee_base_focal.
+The Docker files accept the following version args (note that they must match up):
+
+- `UBUNTU_VERSION` - The version of Ubuntu to use. Valid values are "16.04", "18.04", and "20.04".
+- `ROS_VERSION` - The version of ROS to use. Valid values are "kinetic", "melodic", and "noetic".
+- `PYTHON` - The version of Python to use. Valid values are "" (an empty string representing Python 2) and "3".
+
+If `UBUNTU_VERSION` is `"16.04"`, `ROS_VERSION` and `PYTHON` must be `"kinetic"` and `""` respectively.
+If `UBUNTU_VERSION` is `"18.04"`, `ROS_VERSION` and `PYTHON` must be `"melodic"` and `""` respectively.
+If `UBUNTU_VERSION` is `"20.04"`, `ROS_VERSION` and `PYTHON` must be `"neotic"` and `"3"` respectively.
```
trey0, line 26: noetic

PR #277, opened 2021-09-02
adding service to perching arm to try and reconnect the arm

marinagmoreira:
To reread the arm after fsw has started, one can run:
rosservice call /hw/arm/enable_arm "enabled: true"

This has been tested in the lab


bcoltin (approved):
Add a comment explaining how this can be used if we forget to power the arm and then I think it's good

kbrowne15 (approved):

PR #281, opened 2021-09-09
fix header order

marinagmoreira:
changing the order because it matters for cross compile (failed compiling) ; removing previously included header

### DIFF
--- a/mobility/planner_qp/planner_qp/src/planner_qp.cc
+++ b/mobility/planner_qp/planner_qp/src/planner_qp.cc
@@ -27,6 +27,7 @@
 #include <pluginlib/class_list_macros.h>
 
 // For the planner implementation API
+#include <decomp_util/ellipse_decomp.h>
 #include <choreographer/planner.h>
 
 // Keepout zones for the planner
@@ -36,9 +37,6 @@
 #include <traj_opt_basic/types.h>
 #include <traj_opt_pro/nonlinear_trajectory.h>
 #include <traj_opt_ros/ros_bridge.h>
-
-#include <decomp_util/ellipse_decomp.h>
-#include <jps3d/planner/jps_3d_util.h>
 
 #include <mapper/point_cloud.h>
 #include <pcl/point_types.h>
--- a/mobility/planner_qp/planner_qp/src/planner_qp.cc
+++ b/mobility/planner_qp/planner_qp/src/planner_qp.cc
@@ -28,6 +28,7 @@
 
 // For the planner implementation API
 #include <decomp_util/ellipse_decomp.h>
+#include <jps3d/planner/jps_3d_util.h>
 #include <choreographer/planner.h>
 
 // Keepout zones for the planner


bcoltin (approved):
Can go ahead and merge once it passes tests

trey0:
I don't know all the details of this case, but here are a couple of general guidelines that seem possibly relevant:
- The order of including headers should never matter. If it does matter, usually it is a situation like this: header B implicitly depends on header A but doesn't explicitly include it, so if a source file includes A then B it will compile, but not if it includes B then A. The preferred solution in that case is to modify header B to explicitly include header A, so from that point on the include order no longer matters.
- I saw you removed an include statement. I'm sure you probably did some analysis to figure out it would be ok, and it's probably totally fine. I just want to point out that in general sometimes your code will compile even with an include statement removed, but what you've done is actually to create one of those bad situations like I just mentioned above. A common example is if you have a source file Z that includes a header Y, and Y in turn includes another header X. Now you can use things from header X in the source file Z, but you have become vulnerable to somebody changing the implementation of Y later so it no longer includes X. So it's better to directly include everything you actually use.

All of that said, there may not be any problem with this change, just viewing it as a possible opportunity to improve quality of code we inherited from long ago.

marinagmoreira:
> I don't know all the details of this case, but here are a couple of general guidelines that seem possibly relevant:
> 
> * The order of including headers should never matter. If it does matter, usually it is a situation like this: header B implicitly depends on header A but doesn't explicitly include it, so if a source file includes A then B it will compile, but not if it includes B then A. The preferred solution in that case is to modify header B to explicitly include header A, so from that point on the include order no longer matters.
> * I saw you removed an include statement. I'm sure you probably did some analysis to figure out it would be ok, and it's probably totally fine. I just want to point out that in general sometimes your code will compile even with an include statement removed, but what you've done is actually to create one of those bad situations like I just mentioned above. A common example is if you have a source file Z that includes a header Y, and Y in turn includes another header X. Now you can use things from header X in the source file Z, but you have become vulnerable to somebody changing the implementation of Y later so it no longer includes X. So it's better to directly include everything you actually use.
> 
> All of that said, there may not be any problem with this change, just viewing it as a possible opportunity to improve quality of code we inherited from long ago.

In this case, the header that was giving errors was #include <decomp_util/ellipse_decomp.h>, which is an external dependency. I removed #include <jps3d/planner/jps_3d_util.h>, because it is already included in #include <choreographer/planner.h>

trey0:
> In this case, the header that was giving errors was #include <decomp_util/ellipse_decomp.h>, which is an external dependency (my guess is that those 2 external packages redefine the same values, but there's nothing we can do about it because decomp_util and jps3d are external to astrobee).

Ok, might not be worth tracking down then.

> I removed #include <jps3d/planner/jps_3d_util.h>, because it is already included in #include <choreographer/planner.h>

I see. This looks like an example of the second guideline. I would argue that, because planner_qp.cc does use types defined in jps3d, it is actually better to keep the explicit dependency on jps3d instead of relying on another module to include that header, even if it makes the code a little bit longer.

PR #286, opened 2021-09-21
ubuntu 16/18 linter changes

marinagmoreira:
fixing header for ubuntu 16;
disabling python linter unless in ubuntu 20 - black doesn't work on older python versions


bcoltin:
It would be better to check the python / black version instead of the ubuntu version which isn't the direct issue. Someone could install a newer version on an older ubuntu or vice versa. Could we instead check the python / black version and print a warning if the version is too old?

kbrowne15 (approved):

bcoltin (approved):

PR #293, opened 2021-09-29
Fix error output validator

marinagmoreira:
if destination was not in keepin zones, then it would give the error that the trajectory passed a keepout zone because of the added border between modules, like this, it evaluates the destination points first, outputting the correct error


bcoltin (approved):

trey0:
This doesn't look like valid code for iterating through a std::vector in reverse!

See for example https://stackoverflow.com/questions/3610933/iterating-c-vector-from-the-end-to-the-beginning

bcoltin:
Darn, I like Marina's version of C++ better than the reality :(

PR #294, opened 2021-09-30
using reverse iterator

marinagmoreira:
fixing https://github.com/nasa/astrobee/pull/293


bcoltin (approved):

trey0 (approved):
I agree with Brian, except I think you should be able to just say "for it in reversed(seg):". I don't know why we still have to write for loops like this after *50 years* of "improvements" to the C language.

Thanks, Marina!

PR #295, opened 2021-09-30
Revert gnc visualizer to python2

marinagmoreira:
Even though it works with ros messages in python3, it does not with dds.
Needs further analysis, reverting (after passing it through linter) such that the version in develop fully works.

### DIFF
--- a/tools/gnc_visualizer/scripts/communications/dds_support.py
+++ b/tools/gnc_visualizer/scripts/communications/dds_support.py
@@ -17,45 +17,30 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-from os import path as osPath
-from sys import path as sysPath
-from time import sleep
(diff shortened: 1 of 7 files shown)


tools/gnc_visualizer/scripts/communications/dds_support.py
```diff
@@ -245,18 +245,18 @@ def __dictionary_to_ekf_msg(self, dic):
         for i in range(0, len(dic["cov_diag"])):
             msg.cov_diag[i] = dic["cov_diag"][i]
 
-        msg.confidence = dic["confidence"]
```
bcoltin, line 248: This looks like we accidentally changed the spacing here

bcoltin (changes requested):

tools/gnc_visualizer/scripts/communications/dds_support.py
marinagmoreira: fixed it, thanks for spotting it!

bcoltin (approved):

PR #303, opened 2021-10-18
Debayer live

marinagmoreira:
Adding image_proc nodelet to debayer image
Adding nav cam and dock cam bayer image topics to image_sampler


marinagmoreira:
@trey0 Looks like we're moving towards doing this offline instead, should we close this PR?

trey0:
> @trey0 Looks like we're moving towards doing this offline instead, should we close this PR? The alternative is to make this disabled by default.

I like the idea of keeping it but disabling it by default. The onboard color imagery could help Ian's onboard semantic localization if we are able to slip in an experiment on that. But it should not be a priority if it's going to take significant work.

marinagmoreira:
For this PR to work on the robot, we need to install:
ros-kinetic-image-geometry_1.12.8-0xenial-20210507-111027-0800_armhf.deb
ros-kinetic-image-proc_1.12.23-0xenial-20210507-175115-0800_armhf.deb and

marinagmoreira:
Tested in the robot

hardware/is_camera/src/camera.cc
```diff
@@ -288,18 +294,43 @@ namespace is_camera {
...
+        bayer_img_msg_buffer_[i]->height = kImageHeight;
+
+        // This was tested in the lab using a color test picture
+        // Images in https://github.com/nasa/astrobee/issues/434
+        bayer_img_msg_buffer_[i]->encoding = "bayer_grbg8";
+        bayer_img_msg_buffer_[i]->step = kImageWidth;
+        bayer_img_msg_buffer_[i]->data.resize(kImageWidth * kImageHeight);
+      }
+    } else if (!req.enable && bayer_enable_) {
+      bayer_pub_.shutdown();
```
trey0, line 326: It would be cleaner to call `reset()` on the buffers here as well, to free up some memory. This version is not a memory leak exactly, but it will have some strange behavior if you enable/disable/enable.

astrobee/config/communications/dds_ros_bridge.config
```diff
@@ -198,4 +198,6 @@ cameras = {
   {name="dock_cam", valid_resolutions={"1280_960", "1024_768", "640_480", "320_240"}, mode="FRAMES", max_frame_rate=15},
   {name="sci_cam", valid_resolutions={"1920_1080", "1280_720", "960_540", "480_270"}, mode="VIDEO", max_frame_rate=30},
   {name="haz_cam", valid_resolutions={"224_171"}, mode="FRAMES", max_frame_rate=5},
-  {name="perch_cam", valid_resolutions={"224_171"}, mode="FRAMES", max_frame_rate=5}}
+  {name="perch_cam", valid_resolutions={"224_171"}, mode="FRAMES", max_frame_rate=5},
+  {name="nav_cam_bayer", valid_resolutions={"224_171"}, mode="FRAMES", max_frame_rate=5},
```
bcoltin, line 202: Is this the right thing to do? @kbrowne15 

Not sure we need to send the color images to the ground with dds? And if we do, it would not be the resolution of the perch cam?

shared/ff_util/include/ff_util/ff_names.h
```diff
@@ -382,6 +390,9 @@
 #define TOPIC_HARDWARE_IMU                          "hw/imu"
 #define TOPIC_HARDWARE_NAV_CAM                      "hw/cam_nav"
 #define TOPIC_HARDWARE_DOCK_CAM                     "hw/cam_dock"
+#define TOPIC_HARDWARE_CAM_SUFFIX_BAYER_RAW         "_bayer/image_raw"
+#define TOPIC_HARDWARE_CAM_SUFFIX_BAYER_MONO        "_bayer/image_mono"
```
trey0, line 394: I have mixed feelings about including this topic suffix. Probably, nobody should ever subscribe to `/hw/cam_nav_bayer/image_mono` because its content is essentially the same as `/hw/cam_nav`, but subscribing to it would cause the `debayer` nodelet to redundantly do the same processing that `is_camera` already did.

(Actually, the two versions of debayering would be very subtly different due to the R/B channel swap issue, but that makes it even more of a confusing trap...)

If you really think we should have this here, I would recommend a comment that explains the issue.

astrobee/config/communications/dds_ros_bridge.config
marinagmoreira: Yeah I'm totally down to take out the image_sampler part, and if we ever need it, add it back one day

shared/ff_util/include/ff_util/ff_names.h
```diff
@@ -382,6 +390,9 @@
 #define TOPIC_HARDWARE_IMU                          "hw/imu"
 #define TOPIC_HARDWARE_NAV_CAM                      "hw/cam_nav"
 #define TOPIC_HARDWARE_DOCK_CAM                     "hw/cam_dock"
+#define TOPIC_HARDWARE_CAM_SUFFIX_BAYER_RAW         "_bayer/image_raw"
```
trey0, line 393: I think I understand the logic of why you needed to modify the previous bayer topic by appending `/image_raw` here. 

But there are existing bag recording profiles in other repos that refer to the old name, so we just need to make sure they get updated whenever this change takes effect.

hardware/is_camera/src/camera.cc
```diff
@@ -288,18 +294,43 @@ namespace is_camera {
...
     if (thread_running_) {
       v4l_->SetParameters(camera_gain_, camera_exposure_);
     }
   }
 
+  // Enable or disable the feature timer
+  bool CameraNodelet::EnableBayerService(ff_msgs::SetBool::Request & req,
+                     ff_msgs::SetBool::Response & res) {
+    if (req.enable && !bayer_enable_) {
+      bayer_camera_topic_ = camera_topic_ + TOPIC_HARDWARE_CAM_SUFFIX_BAYER_RAW;
```
trey0, line 311: Right now, bayer variable setup/teardown happens in multiple places, with a lot of code copy/pasted. It makes it harder to understand all the states the variables could be in.

I would recommend creating a single function that can be called both at node initialization and at service enable/disable so that all the setup/teardown goes through the one code path.

hardware/is_camera/src/camera.cc
```diff
@@ -230,6 +233,9 @@ namespace is_camera {
         bayer_img_msg_buffer_[i]->data.resize(kImageWidth * kImageHeight);
       }
     }
+    // Start bayer enable/disable service
+    enable_bayer_srv_ = nh->advertiseService(SERVICE_HARDWARE_BAYER_ENABLE,
```
trey0, line 237: I'm curious about the intended new conops for managing bayer publishing.

The old approach was: If disabled in the config, no bayer publishing. If enabled in the config, publish bayer only if there is a subscriber that wants it. (So there was already a way to turn publishing on/off dynamically by changing recording profiles or whatever, assuming you can control the subscribers.)

What is the intended use case for making dynamic enabe/disable calls to this service? I guess I want to avoid the usability pitfall of setting it to be disabled by default so that users need to dynamically enable it when needed, but they might not realize enabling the bayer images in the image sampler is not enough, and have no good way to discover this extra step.

hardware/is_camera/src/camera.cc
trey0: Ouch. I take it back! I forgot we are using a dangerous shared memory setup and downstream nodelets could still be using the memory in these buffers.

Your code is correct to keep these buffers allocated after disable, but there should also be a guard that prevents them from being reinitialized at the next enable call.

hardware/is_camera/src/camera.cc
marinagmoreira: This is not an extra step, but an OR possibility, please note that if the config is set to enable, no extra action is needed.
In case the fsw was started with bayer disable there is no way of enabling it without having to restart it and manually changing the file. We've had similar issues is ops with things like the arm enable, where if we forgot to do it beforehand, we had to restart the entire sw stack (which we fixed already). This might not be super important for ISAAC activities because they always use bayer recording, but for things like crew minimals and lab testing it will make it much easier.

hardware/is_camera/src/camera.cc
marinagmoreira: Are you worried that we'll disable and then enable it back in less that 1s?

astrobee/config/communications/dds_ros_bridge.config
kbrowne15: Yeah, let's leave this out for now. This is only a small part of what needs to be added to the bridge and I rather add everything at the same time.

hardware/is_camera/src/camera.cc
trey0: Here's my perspective: We are already accepting the risk of using an unsafe memory construct, only because it has a significant performance impact in a place where we need it. I don't see a need to add new potential failure modes on top of that. I think it's safest to (1) leave the buffers in place on a disable call, (2) put a guard on the enable call initialization that checks whether the first buffer is non-null, and if so just skips the buffer initialization loop (to avoid invalidating the existing buffers).

And as always, whenever we do something that looks strange in the code, it would be good to comment explaining why.

By the way, to directly answer your question: I agree that it's unlikely somebody will disable/enable within 1 second, but generally speaking, it's more reliable to use code correctness guarantees that don't depend on assumptions about the speed of things happening. There's no real guarantee about how quickly somebody will send those commands, nor any real guarantee about how long the downstream nodelets will take to process the buffers, and that's especially true if you think about the possibility of future code changes to those nodelets by developers who might not be aware that their image buffer is a time bomb and they have to finish processing it before it explodes. (Explosion = hard-to-debug crash caused by memory corruption.)

hardware/is_camera/src/debayer.cc
```diff
@@ -0,0 +1,130 @@
...
+    if (!pub_color_.getNumSubscribers()) return;
+
+    int bit_depth = enc::bitDepth(raw_msg->encoding);
+
+    if (enc::isBayer(raw_msg->encoding)) {
+      int type = bit_depth == 8 ? CV_8U : CV_16U;
+      const cv::Mat bayer(raw_msg->height, raw_msg->width, CV_MAKETYPE(type, 1),
+                          const_cast<uint8_t*>(&raw_msg->data[0]), raw_msg->step);
+
+        sensor_msgs::ImagePtr color_msg = boost::make_shared<sensor_msgs::Image>();
```
bcoltin, line 83: Since we aren't currently planning to use this, I don't think we need to fix this--- but FYI would probably be easier on the heap to allocate a rotating buffer up front rather than allocating each time

bcoltin (approved):

hardware/is_camera/src/debayer.cc
trey0: Well I agree that luckily we don't need to settle this issue now.

Whenever that discussion comes up, as a preview, my preference would be to avoid more use of the unsafe memory practice (shared_ptr pointing to ring buffer) without some empirical evidence that the performance impact is significant in this case. The color debayer operation might only be invoked at 1 Hz and does not block the critical path for VIO in the navigation control loop.

trey0 (approved):
These changes look great! All of my earlier comments were resolved.

PR #304, opened 2021-10-20
IMU Aug and Heartbeat fixes

rsoussan:
Fixed IMU augmentor to use latest timestamp from latest IMU msg when at standstill, also added logic to prevent IMU Augmentor from sending any repeat msgs with the same timestamp.  Finally, throttled the heartbeat msgs for the IMU augmentor and graph_localizer to publish at max 1 second frequency.  Tested and verified each fix in simulation


marinagmoreira:
I'll test this again this Thursday to merge it in

rsoussan:
Closed as this was merged into updated IMU aug fix pr #341

PR #313, opened 2021-11-02
Added Groundtruth Creation Scripts

rsoussan:
Added scripts to create groundtruth for a bagfile and generate a new groundtruth map and results bagfile.  Also added a helper script to run multiple groundtruth creation processes in parallel.

### DIFF
--- a/tools/graph_bag/readme.md
+++ b/tools/graph_bag/readme.md
@@ -31,3 +31,23 @@ The imu analyzer tool plots imu data and also filtered imu data.  It can use pyt
 
 ## merge\_bags
 Merge bags looks for bag files in the current directory with a provided name prefix and merges these into a single bag file.  Bags are assumed to be numbered and are merged in numerical order.
+
+## run\_graph\_bag\_and\_plot\_results
+Generates localization results for a provided bagfile.
+See 'rosrun graph\_bag run\_graph\_bag\_and\_plot\_results.py -h'
+for further details and usage instructions.
+
+## make\_groundtruth.py
+Generates groundtruth map and bagfile for a given bagfile.
+See 'rosrun graph\_bag make\_groundtruth.py -h'
+for further details and usage instructions.
+
+## make\_groundtruth\_map.py
+Generates groundtruth map for a given bagfile.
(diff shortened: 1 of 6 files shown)


bcoltin (changes requested):
Could you add a comment to the top of each of the scripts explaining what it does and how to use it?

trey0:
> Could you add a comment to the top of each of the scripts explaining what it does and how to use it?

A good pattern is to format the comment as a Python docstring (just a bare string literal at the top of the file, conventionally written with triple quotes) and there is a handy pattern for including the docstring in the argparse help [1].

[1] https://stackoverflow.com/questions/18106327/display-pydocs-description-as-part-of-argparse-help

rsoussan:
> Could you add a comment to the top of each of the scripts explaining what it does and how to use it?

Done using Trey's suggestion

rsoussan:
> Could you add a comment to the top of each of the scripts explaining what it does and how to use it?

Done using Trey's suggestion

bcoltin:
Could you say a bit more? Like, what is a groundtruth map? I still am confused about what this does.

rsoussan:
> Could you say a bit more? Like, what is a groundtruth map? I still am confused about what this does.

Ok good point.  I updated the scripts again, let me know if they need further clarification.

bcoltin (approved):

marinagmoreira:
The documentation in tools/graph_bag/readme.md doesn't mention this.

Unrelated to this, also noticed that there is a repeated merge_bag tool on this node (besides the one in localization/localization_node/tools/merge_bags.cc)

rsoussan:
> The documentation in tools/graph_bag/readme.md doesn't mention this.
> 
> Unrelated to this, also noticed that there is a repeated merge_bag tool on this node (besides the one in localization/localization_node/tools/merge_bags.cc). Oleg converted the old merge_bag tool in the localization_node to c++ because the python api didn't work when we had sci cam images.

Ok I'll update the readme.  The merge_bags tool in graph_bag automatically finds bags in a directory with a given prefix and also optionally filters to only include localization topics.  This is useful to easily combine hivemind bags for localization testing

marinagmoreira (approved):

PR #315, opened 2021-11-03
adding flight enabled different from control enabled

marinagmoreira:
Tested this in the lab by:
 * docking/undocking; 
 * disabled ctl using service, tried to undock - failed to spin up; 
 * enabled ctl using service, undock/dock

### DIFF
--- a/gnc/ctl/src/ctl.cc
+++ b/gnc/ctl/src/ctl.cc
@@ -42,7 +42,7 @@ namespace ctl {  // Nodehandle for <robot>/gnc/wrapper
 Ctl::Ctl(ros::NodeHandle* nh, std::string const& name) :
   fsm_(WAITING, std::bind(&Ctl::UpdateCallback,
     this, std::placeholders::_1, std::placeholders::_2)),
-      name_(name), inertia_received_(false), control_enabled_(false) {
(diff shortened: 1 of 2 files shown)


gnc/ctl/src/ctl.cc
```diff
@@ -510,7 +510,7 @@ bool Ctl::Step(void) {
       NODELET_DEBUG_STREAM_THROTTLE(10, "GNC step waiting for inertia");
       return false;
     }
-    if (!control_enabled_) {
+    if (!flight_enabled_ || !control_enabled_) {
```
marinagmoreira, line 513: please double check this condition to make sure it's sane

marinagmoreira:
@albee please review this too (I can't add you to the reviewers)

gnc/ctl/src/ctl.cc
bcoltin: Looks right to me!

bcoltin (approved):

gnc/ctl/src/ctl.cc
albee: Looks good to me! `control_enabled_` should stay set to the value set by the service this way.

albee (approved):
Looks good to me! Have not tested on hardware but the logic checks out.

PR #323, opened 2021-11-05
Docker remote run

jdekarske:
This adds an option to the run.sh script to use the remote image from github (and downloading it if it doesn't exist) rather than using a locally built one. This could save some hassle from having to build the image locally.

TODO
- [x] add a few sentences to the docs
~- [ ] add a quickstart section to the base readme~ already there under usage

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,16 +2,17 @@
 
 ## Quick Start using the Astrobee docker image
 
-If you just want to try out the astrobee simulator, you can use one of the docker images in the [Docker Hub](https://hub.docker.com/r/astrobee/astrobee). 
+If you just want to try out the astrobee simulator, you can use one of the docker images in the [Github Container Registry](https://github.com/nasa/astrobee/pkgs/container/astrobee).
 
-Given that Docker (see note) is installed:
+Make sure you have docker installed in your ubuntu system following the [installation instructions](https://docs.docker.com/engine/install/ubuntu/) and [post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
 
-	git clone https://github.com/nasa/astrobee.git
-	cd astrobee
-	./scripts/docker/run.sh
+For some systems (with discrete graphics cards), you may need to install additional software. For Nvidia graphics cards, [see here](https://github.com/NVIDIA/nvidia-docker).
 
-*Note: Be aware that this only works on a native ubuntu install.*
-*Note: Make sure you have docker installed in your ubuntu system following the [installation instructions](https://docs.docker.com/engine/install/ubuntu/) and [post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).*
+``` bash
+git clone https://github.com/nasa/astrobee.git
+cd astrobee
+./scripts/docker/run.sh --remote
+```
 
 ## Building the code natively
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,11 +2,13 @@
 
 ## Quick Start using the Astrobee docker image
 
+*The following has been tested on Ubuntu systems using X11 (the default). Please see [these ROS pages](http://wiki.ros.org/docker/Tutorials#Tooling_with_Docker) for more resources.*
+
 If you just want to try out the astrobee simulator, you can use one of the docker images in the [Github Container Registry](https://github.com/nasa/astrobee/pkgs/container/astrobee).
 
 Make sure you have docker installed in your ubuntu system following the [installation instructions](https://docs.docker.com/engine/install/ubuntu/) and [post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
 
-For some systems (with discrete graphics cards), you may need to install additional software. For Nvidia graphics cards, [see here](https://github.com/NVIDIA/nvidia-docker).
+For some systems (with discrete graphics cards), you may need to install [additional software](http://wiki.ros.org/docker/Tutorials/Hardware%20Acceleration).
 
 ``` bash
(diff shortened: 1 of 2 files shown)


trey0:
The documentation change seems mostly good to me, but it also removed the note that the Docker run.sh requires your host OS to be native Ubuntu. Is that not a valid note anymore?

If it makes sense to keep the note, it might be good to move it earlier in the process. (In general, system requirements should come early so people don't start the process then get disappointed partway through when they realize they can't finish.)

It might also make sense to explain a bit why the restrictions are there. Like in theory a user could modify the run.sh script pretty easily to support other Linux flavors, right, might just involve tweaking commands related to X forwarding or something? And the requirement to run the Docker container in a native host OS rather than a VM is related to GPU acceleration? I feel like explaining it is good because one of the main selling points of Docker is that the containers are portable, so it's counter-intuitive that our approach is not portable, but I'm sure it makes sense once you go through it.

jdekarske:
I mostly removed it because I didn't understand the restrictions. I'm not sure there is anything special about Ubuntu which makes it a requirement, as far as Linux distros go. And likely, anyone on a different distro likely has the skillset you mention. This doesn't even preclude someone using WSL, with X11 forwarding.

Basically, in my opinion, native Ubuntu is not a requirement. I agree that it may be informative to change the verbage rather than deleting it.

trey0:
> I mostly removed it because I didn't understand the restrictions. I'm not sure there is anything special about Ubuntu which makes it a requirement, as far as Linux distros go. And likely, anyone on a different distro likely has the skillset you mention. This doesn't even preclude someone using WSL, with X11 forwarding.
> 
> Basically, in my opinion, native Ubuntu is not a requirement. I agree that it may be informative to change the verbage rather than deleting it.

I like how you put it simply in your updated text, especially useful that you link to good additional resources.

If there is anything to change, it might be "Ubuntu" -> "native (non-VM) Ubuntu". (Unfortunately, working in a VM is what all of our interns think of trying first on their locked-down government laptops. If we know it won't work, we should tell them up front.)

jdekarske:
Good point, fixed!

scripts/docker/run.sh
```diff
@@ -59,30 +65,32 @@ touch $XAUTH
...
-docker run -it --rm --name astrobee \
-        --volume=$XSOCK:$XSOCK:rw \
-        --volume=$XAUTH:$XAUTH:rw \
-        --env="XAUTHORITY=${XAUTH}" \
-        --env="DISPLAY" \
-        --gpus all \
-      astrobee/astrobee:latest-ubuntu16.04 \
-    /astrobee_init.sh roslaunch astrobee sim.launch $args
+localtag=astrobee/astrobee:latest-ubuntu16.04
+remotetag=ghcr.io/nasa/astrobee:latest-ubuntu16.04
```
marinagmoreira, line 69: to simplify might be better to just make localtag=astrobee and remotetag=ghcr.io/nasa, and the rest stays as it was instead of defining these variables 3 times each

INSTALL.md
```diff
@@ -2,16 +2,19 @@
...
 
-Given that Docker (see note) is installed:
+If you just want to try out the astrobee simulator, you can use one of the docker images in the [Github Container Registry](https://github.com/nasa/astrobee/pkgs/container/astrobee).
 
-	git clone https://github.com/nasa/astrobee.git
-	cd astrobee
-	./scripts/docker/run.sh
+Make sure you have docker installed in your ubuntu system following the [installation instructions](https://docs.docker.com/engine/install/ubuntu/) and [post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
 
-*Note: Be aware that this only works on a native ubuntu install.*
```
marinagmoreira, line 13: This is still true, right?

scripts/docker/run.sh
```diff
@@ -59,30 +65,32 @@ touch $XAUTH
...
-      astrobee/astrobee:latest-ubuntu18.04 \
-    /astrobee_init.sh roslaunch astrobee sim.launch $args
+localtag=astrobee/astrobee:latest-ubuntu18.04
+remotetag=ghcr.io/nasa/astrobee:latest-ubuntu18.04
 elif [ "$os" = "focal" ]; then
+localtag=astrobee/astrobee:latest-ubuntu20.04
+remotetag=ghcr.io/nasa/astrobee:latest-ubuntu20.04
+fi
+
+# check if docker tag exists
```
marinagmoreira, line 78: it's easier to just define tag as a default in the start and overwrite it if there are arguments defined by the user (-r)

marinagmoreira (changes requested):

INSTALL.md
jdekarske: Trey suggested we move it to the beginning of the section.

scripts/docker/run.sh
jdekarske: Yep, fixed!

scripts/docker/run.sh
```diff
@@ -49,40 +55,35 @@ while [ "$1" != "" ]; do
...
+if [ "$os" = "xenial" ]; then
+tag=astrobee:latest-ubuntu16.04
+elif [ "$os" = "bionic" ]; then
+tag=astrobee:latest-ubuntu18.04
+elif [ "$os" = "focal" ]; then
+tag=astrobee:latest-ubuntu20.04
+fi
+
+# check if local docker tag exists
+if [ "$tagrepo" = "astrobee" ] && [[ "$(docker images -q $tagrepo/$tag 2> /dev/null)" == "" ]]; then
```
marinagmoreira, line 67: I like this!

marinagmoreira (approved):

PR #330, opened 2021-11-10
Fix comparison logic when pushing docker images

rgarciaruiz:
A blank space is needed between the operator and the operands.
Otherwise, it always returns true.


marinagmoreira (approved):

marinagmoreira:
is this PR incomplete to be labelled as draft?

PR #333, opened 2021-11-12
Small fixes for groundtruth generation pipeline

rsoussan:
Fixed some issues that caused faults in the groundtruth pipeline when running multiple bags concurrently.  Also fixed a path issue where a dependent script was pointing to astrobee/astrobee to the config path instead of the updated astrobee/src/astrobee. Finally, added merge_maps from sparse mapping to the cmake file (this might have been an accidental relic of the catkinizing pr).


tools/graph_bag/scripts/utilities.py
```diff
@@ -77,6 +77,7 @@ def load_dataframe(files):
 
 
 def run_command_and_save_output(command, output_filename):
+    print(command)
```
bcoltin, line 80: did you mean to leave this?

bcoltin (approved):

bcoltin:
Looks good once you fix lint (and double check the print is what you wanted)

tools/graph_bag/scripts/utilities.py
rsoussan: Oh yeah I did, but I'll make it optional and pass a bool for printing.  I found it helpful to have the command being run printed to the file but I can see how other times people wouldn't want that

trey0: The logging module may be a good way to handle that.

marinagmoreira (approved):

marinagmoreira (approved):

PR #337, opened 2021-11-16
INSTALL.md: Use spaces for indented blocks to gen correct markup

oleg-alexandrov:
The html currently in https://nasa.github.io/astrobee/html/md_INSTALL.html does not display correctly the "git clone ..." block on top. 

Based on how indented blocks are supposed to be written, per https://github.github.com/gfm/#indented-code-block, I replaced tabs with four spaces. Some other minor fixes were made for consistency of capitalization and punctuation.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,17 +1,17 @@
 # Install
 
-## Quick Start using the Astrobee docker image
+## Quick start using the Astrobee Docker image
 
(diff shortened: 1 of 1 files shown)


marinagmoreira:
just merged a PR that overlaps with this one, can you merge in develop?

marinagmoreira (approved):

PR #340, opened 2021-11-17
Print note that the number of errors may have changed.

oleg-alexandrov:
Once the linter flags the errors, clang-format-8 fixes them, but the old number of errors is printed, before the fix. 

Here I add a note to mention that the user should add and commit the files again to see the updated number of errors. Longer term, our hook needs to be modified to run cpplint second time, but for now this may do.

I also reword the "OBJECTION" notice. I always found it mildly insulting. Fixing thousands of format errors is unpleasant enough, without that ugly note in your face.

### DIFF
--- a/scripts/git/cpplint_repo.py
+++ b/scripts/git/cpplint_repo.py
@@ -36,6 +36,7 @@ def print_objection():
     print("Code formatting errors were found.")
     print("==================================")
 
+
 def main():
     num_errors = 0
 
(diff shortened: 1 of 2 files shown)


marinagmoreira (approved):

bcoltin:
Zack will be sad that you got rid of his beautiful "Objection!" artwork.

oleg-alexandrov:
Back then then I thought Zack introduced the linter as a natural extension
of his mean character. :)

PR #341, opened 2021-11-17
Imu aug fix

rsoussan:
Fixed IMU Augmentor to incrementally add IMU data to latest extrapolated state rather than adding all IMU data since latest graph state message in order to reduce computation time.  Combined with previous PR (see https://github.com/nasa/astrobee/pull/304/files) to include fixes for repeat IMU messages and standstill behavior.


marinagmoreira:
I'll test this in the lab tomorrow :)

localization/imu_augmentor/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ find_package(catkin2 REQUIRED COMPONENTS
 
 catkin_package(
   INCLUDE_DIRS include ${GTSAM_INCLUDE_DIR} ${GLOG_INCLUDE_DIRS} 
-  LIBRARIES ${PROJECT_NAME} ${GLOG_LIBRARIES} ${GTSAM_LIBRARIES} 
+  LIBRARIES ${PROJECT_NAME} ${GLOG_LIBRARIES} ${GTSAM_LIBRARIES} gtsam
```
bcoltin, line 41: This should be part of gtsam libraries?

bcoltin (approved):
Looks good once it's tested

localization/imu_augmentor/CMakeLists.txt
rsoussan: For some reason it's not, not sure what the best way to account for that is but for some packages if I don't include it there I get linker errors

bcoltin: You could make a FindGSTAM in the cmake folder which fixes this. Will leave it to your judgement whether it's worth the effort.

marinagmoreira (approved):
Tested this in the lab, the rate of the ctl and fam are now between 70-80Hz and the CPU on the llp is now much lower (the highest CPU load doesn't go over 70% during movement)

PR #344, opened 2021-11-20
Record with rosbag record

marinagmoreira:
Given that the current data_bagger uses up to 30% of cpu and the "rosbag record" command uses 0%, I changed the astrobee recorder to call the cmd line.
Removed all the threads, making this single thread.

I basically re-did the astrobee_recorder.cc:
 * translates record options into a "rosbag record" cmd;
 * removed the checks that are already done by "rosbag record" -- left only the checkDisk because there are some horror stories on forums of the entire disk being filled and the system unuseable.
 * does a fork->exec to start the recording process (used popen2 to leave the possibility of reading/writing).
 
Added some comments to the data_bagger, added some checks, and made the function order match the h-file.

- [x] Test on the robot


marinagmoreira:
I tested this remotely on bsharp, I recorded the MobNavDockRAW and it was a breeze, still using close to nothin
Screenshot from 2021-11-24 14-28-46
g

management/data_bagger/src/astrobee_recorder.cc
```diff
@@ -72,570 +75,166 @@ using std::vector;
...
-        ROS_WARN("Less than 5 x %s of space free on disk with %s.",
-                  options_.min_space_str.c_str(), bag_.getFileName().c_str());
-        writing_enabled_ = true;
-    } else {
-        writing_enabled_ = true;
-    }
-#endif
-    return true;
-}
+    execl("/bin/sh", "sh", "-c", command, NULL);
```
bcoltin, line 223: Can the user passes some weird string as a topic that would get parsed to evaluate an arbitrary command? Would be better if we could do exec on rosbag record directly without letting things get interpreted by the shell.

bcoltin (changes requested):

kbrowne15 (approved):
Once Brian's comment is addressed, it should be good to merge.

bcoltin (approved):
Thanks!

bcoltin:
We will test this with the performance script and tune some parameters before merging this.

PR #346, opened 2021-11-24
Teleop marker

jdekarske:
This tool is based on the teleop tool in the executive package, but uses an rviz interactive marker instead of CLI. It currently supports (un)docking, movements, and stopping. All commands are preempted by new commands.


tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+    cmd.args.resize(1);
+    cmd.args[0].data_type = ff_msgs::CommandArg::DATA_TYPE_INT;
+    cmd.args[0].i = 1;  // TODO(jdekarske) support other berth
+  }
+
+  cmd_publisher.publish(cmd);
+}
+
+void processFeedback(const InteractiveMarkerFeedbackConstPtr& feedback) {
+  switch (feedback->event_type) {
```
rsoussan, line 155: I think this could just be an if statement since only one case is checked for

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+  int_marker.name = "astrobee_teleop";
+  int_marker.description = "Astrobee Command";
+
+  // insert a box
+  makeBoxControl(int_marker);
+  int_marker.controls[0].interaction_mode = interaction_mode;
+
+  InteractiveMarkerControl control;
+
+  tf2::Quaternion orien(1.0, 0.0, 0.0, 1.0);
```
rsoussan, line 232: Try to avoid uncommon abbreviations, "orientation" would be fine here

rsoussan (changes requested):
Nice work, let me know if you have any questions about the comments

tools/interactive_marker_teleop/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+## catkin specific configuration ##
+###################################
+## The catkin_package macro generates cmake config files for your package
+## Declare things to be passed to dependent projects
+## INCLUDE_DIRS: uncomment this if your package contains header files
+## LIBRARIES: libraries you create in this project that dependent projects also need
+## CATKIN_DEPENDS: catkin_packages dependent projects also need
+## DEPENDS: system dependencies of this project that dependent projects also need
+catkin_package(
+#  INCLUDE_DIRS include
```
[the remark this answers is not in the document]
jdekarske: done

tools/interactive_marker_teleop/package.xml
```diff
@@ -0,0 +1,41 @@
...
+  <exec_depend>roscpp</exec_depend>
+  <exec_depend>interactive_markers</exec_depend>
+  <exec_depend>visualization_msgs</exec_depend>
+  <exec_depend>tf2</exec_depend>
+  <exec_depend>tf2_ros</exec_depend>
+  <exec_depend>tf2_geometry_msgs</exec_depend>
+
+
+  <!-- The export tag contains other, unspecified, tags -->
+  <export>
```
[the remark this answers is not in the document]
jdekarske: done

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <ros/ros.h>
+
+#include <interactive_markers/interactive_marker_server.h>
```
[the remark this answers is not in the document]
jdekarske: done

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+    case visualization_msgs::InteractiveMarkerFeedback::MENU_SELECT:
+      switch (feedback->menu_entry_id) {
+        case 1:  // snap to astrobee
+        {
+          // get astrobee position
+          geometry_msgs::TransformStamped t;
+          t = tfBuffer.lookupTransform("world", "body", ros::Time(0));
+          // set marker pose
+          geometry_msgs::Pose p;
+          p.position.x = t.transform.translation.x;  // Help! is there a better way to do this?
```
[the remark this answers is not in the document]
jdekarske: thanks for the suggestion, but I'll leave it since I only use it once

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
jdekarske: done

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+  // Make the marker
+  server.reset(new interactive_markers::InteractiveMarkerServer("interactive_marker_teleop", "", false));
+  ros::Duration(0.1).sleep();
+  menu_handler.insert("Snap to Astrobee", &processFeedback);
+  menu_handler.insert("Go to Position", &processFeedback);
+  menu_handler.insert("Dock", &processFeedback);
+  menu_handler.insert("Undock", &processFeedback);
+  menu_handler.insert("Stop", &processFeedback);
+
+  // menu_handler.insert("Add to plan", &processFeedback);  // TODO(jdekarske) would be cool to add a new marker for
```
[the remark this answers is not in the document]
jdekarske: Yes, a future commit would add functionality here (and the keepout zone TODO). It's minimally viable for what I'm doing, but if there's interest I'll add these features.

These TODOs are closer to feature suggestions, I guess. Would it be better to remove the TODOs and add a list?

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+#include <tf2_ros/transform_listener.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+
+#include <ff_msgs/CommandStamped.h>
+#include <ff_msgs/CommandConstants.h>
+#include <ff_util/ff_names.h>
+
+#include <string>
+
+using visualization_msgs::InteractiveMarker;
```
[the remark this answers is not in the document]
jdekarske: Good call, fixed

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
jdekarske: Still there because of a TODO, see https://github.com/nasa/astrobee/pull/346#discussion_r776480195

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+          break;
+        }
+        default:
+          break;
+      }
+      break;
+
+      // TODO(jdekarske) check for keep out zones here
+      // case visualization_msgs::InteractiveMarkerFeedback::POSE_UPDATE:
+      //   ROS_INFO_STREAM(s.str() << ": pose changed"
```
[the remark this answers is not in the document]
jdekarske: https://github.com/nasa/astrobee/pull/346#discussion_r776480195

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,299 @@
...
+#include <ff_util/ff_names.h>
+
+#include <string>
+
+using visualization_msgs::InteractiveMarker;
+using visualization_msgs::InteractiveMarkerControl;
+using visualization_msgs::InteractiveMarkerFeedbackConstPtr;
+using visualization_msgs::Marker;
+
+std::shared_ptr<interactive_markers::InteractiveMarkerServer> server;
```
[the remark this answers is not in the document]
jdekarske: done

jdekarske:
@rsoussan Thanks for the feedback! I think we're ready for another check.

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+InteractiveMarkerTeleop::InteractiveMarkerTeleop(ros::NodeHandle& nh) : nh(nh) {
+  // Start tf listener
+  tfListener = std::shared_ptr<tf2_ros::TransformListener>(new tf2_ros::TransformListener(tfBuffer));
+
+  // Make the marker
+  server.reset(new interactive_markers::InteractiveMarkerServer("interactive_marker_teleop", "", false));
+  ros::Duration(0.1).sleep();
+  menu_handler.insert("Snap to Astrobee",
+                      std::bind(&InteractiveMarkerTeleop::processFeedback, this, std::placeholders::_1));
```
jdekarske, line 31: Is this best practice for passing member functions to callbacks here?

tools/interactive_marker_teleop/include/interactive_marker_teleop.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include <ros/ros.h>
+#include <interactive_markers/interactive_marker_server.h>
+#include <interactive_markers/menu_handler.h>
+#include <tf2_ros/transform_listener.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+
+#include <string>
+
+namespace vm = visualization_msgs;
```
rsoussan, line 34: Not very important in this case since nothing will be including this header file other than the .cpp but the general rule of thumb is to put the "namespace a = b;" alias only in the cpp files so other files that include this header don't accidentially use this namespace alias that they might not be aware of.  The unfortunate side effect is not being able to use namespace aliases in headers, so vm:: would be replaced with visualization_msgs::  for this file (but not the cpp).

tools/interactive_marker_teleop/include/interactive_marker_teleop.h
```diff
@@ -0,0 +1,65 @@
...
+
+  void sendMobilityCommand(std::string command, const geometry_msgs::Pose& desired_pose);
+
+  void sendMobilityCommand(std::string command);
+
+  void processFeedback(const vm::InteractiveMarkerFeedbackConstPtr& feedback);
+
+  void make6DofMarker(unsigned int interaction_mode, const geometry_msgs::Pose& position);
+
+ private:
```
rsoussan, line 52: For the c++ google style guide that we follow (for the most part), member variables should end in an underscore

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
rsoussan: Yup looks good to me

tools/interactive_marker_teleop/src/interactive_marker_teleop.cpp
```diff
@@ -16,36 +16,41 @@
...
+  menu_handler.insert("Stop", std::bind(&InteractiveMarkerTeleop::processFeedback, this, std::placeholders::_1));
+
+  // menu_handler.insert("Add to plan", &InteractiveMarkerTeleop::processFeedback, this, std::placeholders::_1);  //
+  // TODO(jdekarske) would be cool to add a new marker for each station menu_handler.insert("Remove from plan",
+  // &InteractiveMarkerTeleop::processFeedback, this, std::placeholders::_1);
 
-ros::Publisher cmd_publisher;
-ros::Subscriber ack_subscriber;
+  geometry_msgs::Pose position;
+  make6DofMarker(vm::InteractiveMarkerControl::MOVE_ROTATE_3D, position);
```
rsoussan, line 43: I think you could remove position from the arguments to this function and default construct the geometry_msgs::Pose in the function, although it looks like its setting the int_marker.pose, so you probably don't need it and int_marker.pose will just be default initialized.  Alternatively, you could add a default arg in the header declaration if this will be used with a non default position arg in the future

rsoussan:
> @rsoussan Thanks for the feedback! I think we're ready for another check.

Nice work, left some more small comments but you can merge when they're addressed if you agree with them

rsoussan (approved):

jdekarske:
@marinagmoreira or @bcoltin Good to merge when you are ready!

PR #351, opened 2021-12-01
Fix: configure.sh correctly appends cmake env var to ./bashrc

albee:
Add: newline before cmake append in configure.sh. configure.sh did not correctly append if ./bashrc did not end in a newline (i.e., command was appended to existing line in ./bashrc, even if it was commented, had a command, etc.)

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -309,8 +309,10 @@ if [ $native_build == 1 ] ; then
     # Add our cmake to paths and bashrc
     grep -qF 'source /opt/ros/'$ros_version'/setup.bash' ~/.bashrc || echo 'source /opt/ros/'$ros_version'/setup.bash' >> ~/.bashrc
     cmake_astrobee_path=`catkin locate -s`/cmake
-    grep -qF ${cmake_astrobee_path} ~/.bashrc \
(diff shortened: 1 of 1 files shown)


marinagmoreira (approved):

albee:
I think we might actually be able to get around messing with the ./bashrc altogether by putting statements we want all packages to use in catkin2Config.cmake, e.g.,
```
# Let cmake know of our additional scripts
list(APPEND CMAKE_MODULE_PATH "/home/albee/workspaces/astrobee-ws-td/src/cmake")
list(APPEND CMAKE_PREFIX_PATH "/home/albee/workspaces/astrobee-ws-td/src/cmake")

# use custom CMake functions
include(CreateLibrary)
include(CreateMsgTargets)
include(CreateRosTestTargets)
include(CreateTestTargets)
include(ExternalProjects)
include(FetchContent)
include(InstallLaunchFiles)
```

marinagmoreira:
That's a chicken and an egg problem xD The Catkin2Config is only found because those statements are in the bashrc.

The choice to not use those macros was made to bring astrobee closer to other ros/catkin projects and ease the learning curve, but it is up for discussion of course.

albee:
You're right, I forgot what directory the config was in :+1: . 

Agreed on using standard CMake functions rather than custom ones, this was definitely a point of confusion for me when first using Astrobee. I think we are a niche case in terms of having a legacy project that needs to upgrade the build system.

bcoltin:
What if we pass a -DCMAKE_PREFIX_PATH in configure.sh based on the relative path to the configure.sh script file?

albee:
@bcoltin Pretty sure that would work too. Can do a PR if you'd prefer to do that and not change the ~./bashrc.

PR #359, opened 2021-12-07
changing picture taking command to new name

marinagmoreira:
The sci cam apk now uses "takePicture"


kbrowne15 (approved):
I added a little bit for the continuous picture taking stuff. Please review it before merging.

PR #360, opened 2021-12-10
Quick docker build

jdekarske:
I added a dockerfile that uses the build intermediates from the previous image to speed up the build.

My local build time is now:
5 minutes and 25.1 seconds!! (with 2 workers AMD Ryzen 7 3800X 2.2 GHz) (GHCR uses 2 workers)

GHCR was:
1h 1m 38s total
42 m 46.7s (catkin build time)

GHCR is now:
27m 40s total (with tests)
8 minutes and 31 sec (catkin build time)

Since develop triggered builds are less frequent, I thought this fit better in the PR workflow.

There is a "cache" action for github actions, but I think that is mainly for dependencies.


marinagmoreira (changes requested):
We actually had initially a very similar setup, but it would break too many times because many PR's need a full rebuild (e.g. if you change message definitions). We can keep the astrobee_quick for personal use, but having it replace the github CI will give us troubles again..

.dockerignore
```diff
@@ -1,4 +1,4 @@
-.git/
+# .git/
```
marinagmoreira, line 1: ?

jdekarske: ah yes, I was going to ask about this. Git history is required to set the timestamps of the files. Is there any reason this directory **shouldn't** be included?

jdekarske:
> We actually had initially a very similar setup, but it would break too many times because many PR's need a full rebuild (e.g. if you change message definitions). We can keep the astrobee_quick for personal use, but having it replace the github CI will give us troubles again..

okay, I will revert the ci

.github/workflows/ci_pr.yml
```diff
@@ -12,12 +12,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
```
marinagmoreira, line 15: This looks like good practice though, we should keep it :)

.dockerignore
marinagmoreira: the initial reason to include the .git in dockerignore is to only compare the files that matter for maximizing cache reusability.
If it actually helps with it then we should comment it out

scripts/docker/astrobee_quick.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY --from=build_cache /src/astrobee/devel /src/astrobee/devel/
+COPY --from=build_cache /src/astrobee/.catkin_tools /src/astrobee/.catkin_tools/
+
+COPY . /src/astrobee/src/
+
+# reset the timestamps of the files to when they were committed, then cakin will only build new stuff
+# https://stackoverflow.com/questions/2458042/restore-a-files-modification-time-in-git
+RUN cd /src/astrobee/src/ \
+    && \
+    for f in $(git ls-tree -r -t --full-name --name-only "HEAD") ; do \
```
KhaledSharif, line 31: This is pretty cool, didn't know about this. I will include it in my future ROS+Docker builds.
One question: if you added .git to the .dockerignore, how does this bash command still work?

.github/workflows/ci_pr.yml
jdekarske: it does come with the consequence of ~30s to download the whole history. But that is peanuts compared to the whole build time.

scripts/docker/astrobee_quick.Dockerfile
jdekarske: it's the opposite: _.git must not be ignored_

marinagmoreira (approved):

PR #362, opened 2021-12-14
Documentation failed to compile

marinagmoreira:
I fixed the documentation replacing \subpage with \ref such that it could compile.
\subpage can only be used one for establishing a tree; \ref is used to reference a page

@oleg-alexandrov please re-arrange the tree in this PR if you don't like this configuration :)

@rsoussan reminder to put new documentation (calibration) inside a tree or else it will appear in the first page of the documentation and not under tools. If it was intended, please revert my changes

### DIFF
--- a/astrobee.doxyfile
+++ b/astrobee.doxyfile
@@ -789,7 +789,7 @@ INPUT                  =  README.md \
                           astrobee \
                           simulation \
                           tools \
-                          description \
+                          description/description \
                           licenses.csv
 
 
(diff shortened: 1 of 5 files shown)


bcoltin (approved):

oleg-alexandrov:
Looks good. Based on the gen html I had assumed \subpage makes links, which it does, and I verified that the links were usable, but yeah, I guess that was the wrong way of doing things.

marinagmoreira:
> Looks good. Based on the gen html I had assumed \subpage makes links, which it does, and I verified that the links were usable, but yeah, I guess that was the wrong way of doing things.

It's not a problem if you do \subpage the first time and \ref the others (given that you want the page to be a child of the referring page). The way it was, the doxygen would freeze and never compile because it couldn't parse the tree.

PR #363, opened 2021-12-14
Framestore can now be called multiple times from different workspaces

marinagmoreira:
This is useful for projects like isaac that had specific tf definitions (cargo->cargo_approach or body->soundsee)
Instead of manually specifying these transforms, now we can just have a config in the specific workspace and start a local framestore node.


marinagmoreira:
I had to change the initial PR because the framestore nodelet can't be reused due to it having a hardcoded name/ being linked to the manager.
The existing tool only allows global transforms (in case we start the sim without any robots)
So my solution was to just add a custom transforms tool which allows robot transforms and gets the platform from the node handle instead of through the ff_nodelet

marinagmoreira:
This was tested in sim for astrobee and astrobee+isaac

bcoltin:
Can we get rid of the old framestore then? Would be better to only have one. I see the old one is a nodelet though, with heartbeat etc. Seems like we would want that.

Also, can we pick a more descriptive name than custom_transforms? Maybe broadcast_static_tfs or something along those lines?

bcoltin:
Oh I just read your comment. We had a similar problem with the flashlight nodes since there is one in front and one in back--- try searching the repo for "flashlight_front". I think we can use that to have multiple with different names.

marinagmoreira:
> Oh I just read your comment. We had a similar problem with the flashlight nodes since there is one in front and one in back--- try searching the repo for "flashlight_front". I think we can use that to have multiple with different names.

In the flashlights fix, it was only possible because it is a gazebo plugin and inherits FreeFlyerModelPlugin. I edited it such that if the name was empty I could read it from the urdf. I'm guessing that for a generalized solution we can change the ff_nodelet to use the node name that comes from the NodeHandle - that would be a big change.

marinagmoreira:
> Can we get rid of the old framestore then? Would be better to only have one. I see the old one is a nodelet though, with heartbeat etc. Seems like we would want that.
> 
> Also, can we pick a more descriptive name than custom_transforms? Maybe broadcast_static_tfs or something along those lines?

The framestore nodelet checks the config file every 1s instead of just publishing it and dying.

I didn't give much thought to the name, the other tool is called global_transforms, I can def change it. Do you think we should give a more descriptive name to just this tool or the entire ros node?

bcoltin:
> > Can we get rid of the old framestore then? Would be better to only have one. I see the old one is a nodelet though, with heartbeat etc. Seems like we would want that.
> > Also, can we pick a more descriptive name than custom_transforms? Maybe broadcast_static_tfs or something along those lines?
> 
> The framestore nodelet checks the config file every 1s instead of just publishing it and dying.
> 
> I didn't give much thought to the name, the other tool is called global_transforms, I can def change it. Do you think we should give a more descriptive name to just this tool or the entire ros node?

To clarify, it checks if the config has been modified and if it has re-reads it. Also, your nodelet doesn't publish then die--- it has a ros::spin() at the end. It needs to keep running so people can read the latched tf message later. I don't think the config checking is an important difference that we need a second program for--- let's just always check if the files are modified (or can make a flag if we want).

Actually, now I am more confused, since as you point out there is a global_transforms too! And framestore nodelet and now this new one. What do they do differently? Can we merge what we need from each of them into the nodelet?

marinagmoreira:
> 

That's true, it does run continuously. What we have right now is a framestore nodelet that starts with each robot. The global_transforms starts with the simulation and only publishes global transforms in the case where a simulation is started without any robots.

When I was talking about the flashlights I actually meant https://github.com/nasa/astrobee/blob/master/simulation/src/gazebo_model_plugin_flashlight/gazebo_model_plugin_flashlight.cc#L40, wasn't aware of that one.

bcoltin (approved):

PR #365, opened 2021-12-14
Fix rosdep py dep

Pedro-Roque:
Missing python3-rosdep dependency would return `sudo: rosdep: command not found` during installation. Fixed by adding the dependency `python3-rosdep` to the packages list (added to Focal, might be needed on Xenial and Bionic as well...).


bcoltin:
Can you put this in packages_desktop instead? packages_base is for the robot and I'd rather not change our image.

Pedro-Roque:
Done @bcoltin !

bcoltin:
Thanks!

bcoltin (approved):

PR #367, opened 2021-12-14
Minor fix in bsharp2 config file and robustness in bag extraction logic

oleg-alexandrov:
I found some fragile logic in extract_image_bag.cc which works well in this case but is dangerous in general and resulted in a bug in a different context in isaac. 

When an image is pulled from a bag, it is allocated temporary storage and the cv::Mat object created by the user only points to that storage. That storage can go out of scope any time, then the user cv::Mat is left with junk. An explicit copy out of the temporary storage is what fixes it.

This logic is used in some other places in Astrobee but to put this fix there would likely impact the performance, and it appears that if the image object is used only short-term, the problem does not show up.

In this pull request I also make a minor update to the bsharp2 config file which is only for the geometry mapper and which was validated based on the latest lab data.


bcoltin:
I don't understand, this change is in the main function of extract_image_bag, that shouldn't be used anywhere except in extract_image_bag? The result of toCvShare will be valid while image_msg is valid, which it should always be in this case. This will make the program slower and I don't see the beneift. Also, if you need to make a copy, you can use cv_bridge::toImageCopy

oleg-alexandrov:
The problem, as I see it, is that the previous code was bad design, and if copied thoughtlessly in other code it will result in problems. Which is what I ran into. 

I undid now the change but at least put a comment.

bcoltin:
Thanks!

bcoltin (approved):

PR #368, opened 2021-12-16
github test with single job

marinagmoreira:
lately we've been having some tests fail because github uses multiple jobs on the $catkin make test. This assures that only one test runs at a time


bcoltin (approved):

bcoltin (commented):
Actualy, why do we do catkin build --make-args tests twice?

PR #370, opened 2021-12-21
CPU node load info + Performance scripts

marinagmoreira:
Example of a granite_square analysis as of now:
Screenshot from 2021-12-20 17-55-26

### DIFF
--- a/management/cpu_mem_monitor/src/cpu_mem_monitor.cc
+++ b/management/cpu_mem_monitor/src/cpu_mem_monitor.cc
@@ -507,7 +507,7 @@ int CpuMemMonitor::CollectCPUStats() {
 
     if (sscanf(buffer, "%*d %*s %*c %*d"                          // pid, command, state, ppid
                        "%*d %*d %*d %*d %*u %*u %*u %*u %*u"
-                       "%lu %lu"                                  // usertime, systemtime
+                       "%" SCNu64 " %" SCNu64                     // usertime, systemtime
                        "%*d %*d %*d %*d %*d %*d %*u"
                        "%*u",                                    // virtual memory size in bytes
                        &utime, &stime) != 2)
--- a/management/cpu_mem_monitor/src/cpu_mem_monitor.cc
+++ b/management/cpu_mem_monitor/src/cpu_mem_monitor.cc
@@ -78,7 +78,7 @@ void CpuMemMonitor::Initialize(ros::NodeHandle *nh) {
                             true,
                             false);
 
-  // Timer for clearing the cpu load too high fault
+  // Timer for clearing the memory load too high fault
   clear_mem_load_fault_timer_ = nh->createTimer(
                               ros::Duration(clear_load_high_fault_timeout_sec_),
                               &CpuMemMonitor::ClearMemLoadHighFaultCallback,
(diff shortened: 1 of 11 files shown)


marinagmoreira:
@bcoltin @rsoussan @kbrowne15 What else would be useful to visualize to evaluate/quantify performance?

bcoltin:
> @bcoltin @rsoussan @kbrowne15 What else would be useful to visualize to evaluate/quantify performance?

I think the bottom right "node CPU loads" style is by far the best, we probably don't need the other ones with the same info. Can we add other nodes in so we get close to 400 percent? It would be nice to have a similar graph to this one for LLP. (what is taking all the CPU besides IMU augmentor?)

For the CPU load graphs at the top, I would scale them from 0-100 so we aren't distracted from the tiny drops. Actually, might be nice to do these in the same style as the bottom right with them on top of each other.

I would change the histogram y-axis to percentage instead of count. And I would reorder them in the pipeline order: imu --> ekf --> command. Not sure the frequency graphs to the left of the histograms help much.

It could also be helpful to have a graph of memory usage, especially as we deploy bigger maps.

Thanks, this is looking great! I already have some questions about what on earth is going on with the sparse mapping rates...

marinagmoreira:
The performance scripts are not fully feature complete, but we can do a subsequent PR for this

communications/ff_msgs/msg/CpuStateStamped.msg
```diff
@@ -45,3 +45,5 @@ float32 temp
 # or not all of them are enabled.
 ff_msgs/CpuState[] cpus
 
+# Load usage of individual ROS nodes
+ff_msgs/CpuNodeState[] load_nodes
```
trey0, line 49: I would highly recommend we avoid lightly changing message schemas, because of the problems we discussed before with losing the ability to load data from older bags.

In this case with adding a new field, I would recommend either:
1. Leave the old message schema alone and publish the new field as a separate new message. Old bags would have only the old message and new bags would have both the old and new messages.
2. Leave the old message schema alone and define a new extended message schema as you have it here, but make sure both the new message schema name and the topic it is published on are different (e.g. CpuStateStamped2). Old bags would have only the old message and new bags would have only the new extended message. Analysis tools could still load old bags because the old message definition would still be present, although deprecated for future use.

In the rare cases when you think it's worthwhile to propose changing a message schema, I think we should have a rule to highlight that clearly in the PR summary so it can't accidentally slip through the review process.

marinagmoreira: I think option 2 might be better because option 1 would imply updating all the recording profiles

marinagmoreira: For this specific case I confirmed with Brian that the message was ok to change and Katie confirmed that the bridge is not altered.
The performance script that is a part of the PR that analyses the message will work with both old and new definitions given that we added to the message (uses python) instead of modifying it.

trey0: It sounds like you're saying that the Python bag reader is able to read messages from old bags where the messages on the topic of interest have a missing field according to the current (updated) message schema. That's really surprising. (How does it fill in the value of the missing field?)

Really glad to hear that works!

management/cpu_mem_monitor/src/cpu_mem_monitor.cc
```diff
@@ -461,6 +480,56 @@ int CpuMemMonitor::CollectCPUStats() {
...
 
+  // Go through all the node list and calculate cpu usage
+  uint64_t utime, stime;
+  double cpu_usage_node;
+  std::map<std::string, int>::iterator it;
+  std::map<std::string, uint64_t>::iterator it_load;
+  int i = -1;
+  for ( it = nodes_pid_.begin(); it != nodes_pid_.end(); it++ ) {
+    // Increment counter
+    i++;
```
bcoltin, line 491: A more elegant way to do this is to put in the for loop, for (int it = -1, std::...::iterator it = ... ; ... ; i++, it++) with the comma to do multiple things

tools/performance_tester/readme.md
```diff
@@ -0,0 +1,4 @@
+\page performance_tester Performance Tester
+
+## Usage
+TBD
```
bcoltin, line 4: Can you add a few sentences about what these scripts are and how to use them

bcoltin (approved):
Add a few sentences of documentation for the performance tester and then it's good to merge

PR #373, opened 2021-12-22
Changes to bot config files and some minor fixes

oleg-alexandrov:
This commit changes all the robot config files to reflect a more generic approach of defining relationships between camera transforms. All the camera transforms and timestamp offsets are now relative to nav cam. This is only relevant for ISAAC.

I also put a fix in reading config files so it does not complain when a table is read instead of number for the distortion parameters.  For that, I added to config_reader a function which can peek and see if a field is a number or a table.

I modified the calibration doc to note that nowadays pico_proxy is launched automatically (as long as drivers are launched), but kept the original logic for how to launch it separately if necessary.

Then I fixed a bug in iss.rviz. The sci cam topic should be compressed, otherwise it can't be displayed.


astrobee/config/robots/bsharp.config
```diff
@@ -24,8 +24,15 @@ robot_imu_drdy_pin = 4
 
 robot_geometry = {
 -- Placeholder transforms, not accurate!
-  hazcam_to_navcam_transform = transform(vec3(0.079356943, 0.00019413532, -0.0019157775), quat4(0.076822296, -0.009733734, 0.99698419, 0.0051110388)),
-  scicam_to_hazcam_transform = transform(vec3(0.0070938664, 0.061024816, -0.11184277), quat4(0.065639737, -0.031237584, 0.99720414, -0.017307023)),
+  nav_cam_to_haz_cam_transform = transform(vec3(0.071421481, -0.00030319673, 0.0018058249), quat4(-0.0030431141, 0.0092646368, 0.99993195, 0.0064039206)),
+  nav_cam_to_sci_cam_transform = transform(vec3(-0.076526224, 0.011869553, 0.045409155), quat4(-0.0035414316, 0.0048089362, -0.0071515076, 0.99995659)),
+  nav_cam_to_haz_cam_timestamp_offset = -0.02,
```
bcoltin, line 29: I don't think the geometry section is the best place for time offsets. Maybe in the calibrations section would be better? Or if you have another idea.

bcoltin (changes requested):

oleg-alexandrov:
Strictly speaking the camera transforms are also calibrated quantities, just like the offsets. But  I see your point. The offsets would probably belong under each camera's intrinsics, further down the config file, in the robot_calibrations subsection. 

I'll implement this tomorrow if there's no further comments.

bcoltin:
> Strictly speaking the camera transforms are also calibrated quantities, just like the offsets. But I see your point. The offsets would probably belong under each camera's intrinsics, further down the config file, in the robot_calibrations subsection.
> 
> I'll implement this tomorrow if there's no further comments.

Otherwise it looks good to me! Thanks!

oleg-alexandrov:
I moved the offsets to the robot_calibrations namespace. I checked that the bot files for bumble, honey, queen, and bsharp load well and work as expected, and that the rest follow the same logic (a script was used to consistently convert their format). 

I will merge this once the tests pass.

bcoltin (approved):

PR #374, opened 2021-12-23
Added sparse mapping launch script

rsoussan:
Runs the sparse mapping matcher (localization_node) live using an input bagfile and optionally saves the results to a bagfile.  Created for @oleg-alexandrov for use with the ISAAC pipeline.

### DIFF
--- a/astrobee/launch/offline_localization/record_localization.launch
+++ b/astrobee/launch/offline_localization/record_localization.launch
@@ -18,5 +18,5 @@
 <launch>
   <arg name="output_bagfile"/>             
   <arg name="terminal" default="" />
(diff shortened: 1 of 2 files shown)


bcoltin (approved):
Maybe pass the terminal as an argument to the launch file and default to a more standard one than terminator. Otherwise looks good.

oleg-alexandrov:
I pulled this in, I tested that it works, except for the last line, 

node pkg="rosservice" type="rosservice" name="enable_sparse_mapping" args="call --wait /loc/ml/enable true" 

failing.

oleg-alexandrov:
I think I am very bad at dealing with pull requests. I merged this in prematurely, the clicked on "revert" to see what it will do, which just created a branch, which I plan to wipe.

I think there's not much harm I pulled this in too early. I will keep an eye on the unresolved issue above which shows up only on   ROS melodic and not with ROS kinetic.

bcoltin:
It's fine, just make a new pull request when you fix that problem.

oleg-alexandrov:
I will look at this this week.

PR #380, opened 2021-12-31
make image names uniform; fix debian build script

marinagmoreira:
* Making image names uniform
 * Fixing base image build when added the miro and soracore debian check


bcoltin (approved):
Did you mean to update android? If you did it looks good

KhaledSharif:
I tested this and it works well (I was able to build Astrobee and ISAAC Docker images). It still complains about the IMU bias config thing though.

PR #381, opened 2022-01-01
Added Depth Odometry Sweep Tool and Optional ICP-based refinement for Image Feature Depth Odometry

rsoussan:
Currently the image feature depth odometry is not as accurate as ICP but is much much faster, while ICP is too slow for online usage.  To improve the image-feature estimate, this pr enables passing the estimate from the fast image-feature approach to ICP as an initial guess and ICP is limited to 1 iteration to avoid adding too much runtime.  This hybrid approach is faster than ICP on its own as the iterations are limited and it helps remove some of the more costly ICP situations which tend to occur as the point clouds are further away from each other since an initial guess is provided by the image feature approach.  

Future work includes taking advantage of the fact that our point clouds are organized (have an ordering and height and width since they come from a time-of-flight sensor) to use faster methods for downsampling and normal estimation to further reduce ICP computation time.


bcoltin (approved):

rsoussan:
Closed as this was merged into new depth odometry pr.

PR #383, opened 2022-01-03
Minor config fixes

oleg-alexandrov:
This pull request removes some outdated and obvious comment in cameras.config, and provides updated bumble and bsharp2 calibration parameters with the latest code which perform better than before (only things relevant to isaac change, in particular the bumble nav cam calibration is kept fixed, and the bsharp2 config file is a copy of the bsharp config while used for isaac experiments).


oleg-alexandrov:
I added one more commit fixing some indentation issues.

bcoltin (approved):

PR #388, opened 2022-01-06
Depth odometry organized icp

rsoussan:
Added functions to take advantage of the layout of our point cloud being "organized" (having a width and height since it's from a depth camera) to speed up normal estimation.  Also added functions for sampling using normal alignments to further speed up ICP.  

This branch is off of the open pr " Added Depth Odometry Sweep Tool and Optional ICP-based refinement for Image Feature Depth Odometry #381, so depending on the merge order that can be closed or merged before this one (I couldn't find an option to make the destination that pr, let me know if there's a better way to do this).
This requires new debs to be installed (libpcl-registration and libpcl-keypoints).


bcoltin (approved):
Make sure you add new packages to platform too

PR #389, opened 2022-01-06
Making mapper nice 10

marinagmoreira:
Tested on the robot, output is still very precisely 1Hz (confirmed with htop that nice was aplied)
Performance results from this: https://babelfish.arc.nasa.gov/confluence/display/astrosoft/Astrobee+Profiling


kbrowne15 (approved):
You can merge this once the tests pass.

PR #390, opened 2022-01-07
build_map.md: Update the doc and mention Ryan's tool

oleg-alexandrov:
This is a minor doc adjustment to reflect that when localization_node is used, the bag must be played first, and localization enabled later. 

Also how to use Ryan's tool for these steps is shown, and a gotcha is outlined, which is that the path to the bag must be absolute with this tool, unlike when one plays a bag directly with rosbag play. 

Note that there is bit of repetition in the doc, that is because it describes how localization should work on the real robot and on a local machine.


bcoltin (approved):
Thanks!

PR #393, opened 2022-01-11
Bugfix in undistort_image

oleg-alexandrov:
This commit fixes a couple of issues in undistort_image. When the distortion model has large inaccuracies at the boundary, or when the user would specify large crop dimensions for the undistorted images, it would crash. The solution is to do adjustments to keep things within bounds.

A second fix is to nvm_visualize. This has some functionality for deleting images, which only I use (it is useful though), and sometimes due to how getKey() works in OpenCV the tool thinks the 'Del' key is received when in fact another signal was sent.  The reason for that is obscure. This fix allow images to be deleted in nvm_visualize only with the 'x' key, and a comment about that was added.

### DIFF
--- a/localization/camera/tools/undistort_image.cc
+++ b/localization/camera/tools/undistort_image.cc
@@ -128,17 +128,51 @@ int main(int argc, char ** argv) {
   cv::Mat floating_remap, fixed_map, interp_map;
   cam_params.GenerateRemapMaps(&floating_remap, FLAGS_scale);
 
-  // Deal with the fact that this map may request pixels from the
-  // image that are out of bounds. Hence we will need to grow the
-  // image before interpolating into it.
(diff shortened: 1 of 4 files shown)


oleg-alexandrov:
This commit fixes a couple of issues in undistort_image. When the distortion model has large inaccuracies at the boundary, or when the user would specify a large crop dimensions for the undistorted images, it would crash. The solution is to do adjustments to keep things within bounds.

A second fix is to nvm_visualize. This has some functionality for deleting images, which only I use (it is useful though), and sometimes due to how getKey() works in OpenCV the tool thinks the 'Del' key is received when in fact another signal was sent.  The reason for that is obscure. This fix allow images to be deleted in nvm_visualize only with the 'x' key, and a comment about that was added.

bcoltin (approved):

PR #406, opened 2022-01-28
adding changelog for debians with different dependencies among distros

marinagmoreira:


rgarciaruiz (changes requested):
These look great, but I noticed there's another dependency I believe will require new versions for libalvar2, libdbow21, libdbowdlib1, libdecomputil0, and libopenmvg1: libgcc-s1. I believe that one is new to focal and replaces libgcc1.

Soracore and miro may also need new versions because of libboost. It's looking like we are going to need separate versions for every distribution after all (except opencv).

I'll send you the info in a separate conversation.

bcoltin:
I am a bit confused by this... In the pull request it seems we are only modifying the changelogs. That shouldn't affect the dependencies? If we just need different versions compiled for 16.04 and 20.04, we should compile in each system, and put the appropriate debians in the right folder on astrobee.ndc (maybe I am misunderstanding).

It sounds like we will just need to compile new versions in the 20.04 arm cross compile chroot?

rgarciaruiz:
(Just to finish this conversation, since we are having another on the side...)
When building on every distribution, control dependencies will be expanded to different packages. If we keep the same name, the files will have different checksums and that will break the repo pool. We basically need different names for different distributions.

PR #410, opened 2022-02-03
making the mapper wait for depth instead of reading all

marinagmoreira:
This should reduce the processing time and still produce the 1Hz desired rate (tested in sim)
Will test this in the lab tomorrow.


astrobee/launch/robot/MLP.launch
```diff
@@ -38,7 +38,7 @@
     <node pkg="nodelet" type="nodelet" args="manager"
           name="mlp_depth_cam" output="$(arg output)"/>
     <node pkg="nodelet" type="nodelet" args="manager"
-          name="mlp_mapper" output="$(arg output)" launch-prefix="nice -n 10"/>
+          name="mlp_mapper" output="$(arg output)" launch-prefix="nice -n 5"/>
```
marinagmoreira, line 41: lowered this value due to the reports that the previous one would result in heartbeat and crashing problems

bcoltin (approved):
Looks good once tested

marinagmoreira:
Tested in the lab, mobility improved greatly! The recording also didn't have any gaps as observed before. CPU performance didn't have a major improvement from a visual inspection.

PR #411, opened 2022-02-03
Add Python tools for processing Pico Flexx data, reducing redundant logging requirements

trey0:
These tools allow us to process Pico Flexx "extended" messages to reconstruct "points" point clouds and "amplitude_int" amplitude images.

The intent is to eventually stop logging the latter message topics because they are redundant and take a lot of data volume, but before we can move forward with that, we should double-check that the reconstructed output messages are compatible with existing uses, including ISAAC dense mapping and depth-based localization.

### DIFF
--- a/hardware/pico_driver/scripts/README.md
+++ b/hardware/pico_driver/scripts/README.md
@@ -5,7 +5,8 @@ These Python utilities enable us to process logged
 `/hw/depth_haz/extended` topic and split the data to regenerate both
 point cloud messages normally logged on the `/hw/depth_haz/points` topic
 and amplitude image messages logged on the
-`/hw/depth_haz/extended/amplitude_int` topic.
+`/hw/depth_haz/extended/amplitude_int` topic. (Those are the
+HazCam-specific topics; to use  PerchCam, specify `--cam=perch`.)
 
 As a result, once these tools are fully validated, we will no longer
 need to log the latter two topics, providing a major savings in data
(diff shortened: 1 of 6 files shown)


oleg-alexandrov (approved):
Looks good. A test should be done in the lab, where both the inputs and outputs for this tool are recorded with the bot as they are, then this tool should be run on the inputs, create new outputs, and both the old and new outputs need to be run through the geometry mapper and streaming mapper. 

Special care needs to be given to msg.header.stamp, as this needs to be preserved. Note that msg.header.stamp is not the same as mgs's own timestamp, the latter is publication time/regeneration time rather than creation time. 

If it is ever desired have the bot pick every n-th message, as nav_cam does now, that would be a big change too, and note that the depth cloud field of view is smaller than nav cams, so maybe there's no room for excluding shots. That would be a different discussion.

trey0:
> Looks good. A test should be done in the lab, where both the inputs and outputs for this tool are recorded with the bot as they are, then this tool should be run on the inputs, create new outputs, and both the old and new outputs need to be run through the geometry mapper and streaming mapper.

I agree that testing is important, but I don't see a need to collect new data in the lab. Since the tools are only doing post-processing after the activity, we should be able to do all the testing we need using prior data sets both from the ISS and from the lab. In order to be suitable to use for validation, the prior data set must contain all three relevant depth message topics ("extended", "points", "extended/amplitude_int"), which is typical for ISAAC survey activities.

If you want to pass the output to other tools like the mappers, the process would look something like this:

    $HOME/astrobee/src/hardware/pico_driver/scripts/pico_split_extended.py input.bag haz_split.bag
    $HOME/astrobee/src/scripts/postprocessing/rosbag_topic_filter.py input.bag -r /hw/depth_haz/points -r /hw/depth_haz/extended/amplitude_int input_minus_haz_split.bag
    merge_bags input_minus_haz_split.bag haz_split.bag -o merged.bag
    # run downstream tests on merged.bag

> Special care needs to be given to msg.header.stamp, as this needs to be preserved. Note that msg.header.stamp is not the same as mgs's own timestamp, the latter is publication time/regeneration time rather than creation time.

Each message output by `pico_split_extended.py` is generated from an `extended` message. The complete header of the source extended message is copied to the output message verbatim, so the internal timestamp within the source message is preserved. Next, in the `rosbag.write()` call, `msg.header.stamp` is copied again to be rosbag's external timestamp for that message, which may be used by the rosbag merge scripts for message ordering. Hopefully this behavior will minimize problems for the downstream tools.

A convenient side-effect of this timestamp approach is that the points message and amplitude_int message from the same sensor frame will have guaranteed identical timestamps that could be used for reliable matching. (A weakness of our current onboard logging approach that it's hard to reliably match messages that come from the same frame.)

A topic we should discuss further on another thread is improving the quality of the source timestamps logged onboard the robot to better represent the sensor's true frame acquisition time, and also make matching easier.

oleg-alexandrov:
I see. I did not realize the raw /extended topic is also recorded, on top of the derived results.

rsoussan:
Awesome, glad logging will be more reasonable now!  Looks good from my side, as long as I can generate bags with the original topics from recorded data all my offline tools will work

rsoussan (approved):

Issue #412, opened 2022-02-03
Sensor data message timestamps should represent data acquisition time


trey0:
Astrobee FSW generates many sensor data messages with timestamps. For best data quality, it can be important for these timestamps to reflect the sensor data acquisition time as accurately as possible.

For example, you may want to register the location of a HazCam image in a 3D map using localization from bracketing NavCam images, so the best estimate of the HazCam position comes from interpolating between the NavCam positions. But that means any error in the HazCam timestamp will directly affect its estimated position and reduce the quality of the resulting map.

A typical pattern we have been using is to receive data from a sensor driver, do some of our own minor processing to reformat the data as a ROS message, then publish the ROS message with `msg.header.stamp` set to `ros::Time::now()`. In other words, the timestamp is set to the publish time rather than the data acquisition time, and these times may differ significantly, with no guarantee that the offset between them is consistent (since the processing time can vary based on system load and other factors).

Since we generally have only one timestamp field available in these messages, and there is an important use case that requires the data acquisition time, I argue we should change our publish code so that `msg.header.stamp` is set based on the most accurate data acquisition time we can get:

- In the best case, when the data arrives from the sensor, we may be able to propagate an upstream timestamp recorded by the sensor itself, or by the sensor driver. This can eliminate the effect of upstream latency within the sensor, communication, or sensor driver before the data is received by our code.
- If no upstream timestamp is available, we should at least make the call to `ros::Time::now()` as soon as the data arrives, before we do any further processing.

Furthermore, in some cases, it may make sense to continue propagating the data acquisition time to messages derived from the original sensor data message in downstream nodes. That could be helpful if:

- The original message is not logged, so the downstream message is the only place to find the data acquisition time during post-activity data analysis.
- For some types of data analysis, it may be important to reliably match up multiple messages that were derived from the same sensor frame, in which case it could be very helpful if the messages were guaranteed to have identical timestamps. (Example: the dense mapper needs to process matching point cloud and amplitude_int messages derived from the same HazCam sensor frame.)

Cons of this approach and alternatives:

[...]

PR #413, opened 2022-02-03
Use data acquisition timestamps for Pico sensor data per #412

trey0:
This hasn't been tested yet. The modifications should improve the Pico Flexx sensor data message timestamp accuracy a little.

One case that has some extra uncertainty is the DepthImage timestamp units, since it's not documented. I assumed it uses usecs like DepthData, but if not, it will return bizarre garbage timestamps, so it should be obvious. I think this message is not usually published, at least during ISAAC surveys.

### DIFF
--- a/hardware/pico_driver/src/pico_driver.cc
+++ b/hardware/pico_driver/src/pico_driver.cc
@@ -306,7 +306,7 @@ class PicoDriverL1 : public PicoDriver, public royale::IDepthDataListener, publi
     }
     // If we have depth data, use the same mechanism as L1 to push it
     if (pub_cloud_.getNumSubscribers() > 0) {
-      cloud_.header.stamp = ros::Time::now();
+      cloud_.header.stamp.fromNSec(std::chrono::duration_cast<std::chrono::nanoseconds>(data->timeStamp).count());
       std::copy(
         reinterpret_cast<const uint8_t*>(data->points.data()),
         reinterpret_cast<const uint8_t*>(data->points.data()) + cloud_.row_step * cloud_.height,
@@ -328,7 +328,9 @@ class PicoDriverL1 : public PicoDriver, public royale::IDepthDataListener, publi
     }
     // If we have depth data, use the same mechanism as L1 to push it
     if (pub_depth_image_.getNumSubscribers() > 0) {
-      depth_image_.header.stamp = ros::Time::now();
+      // units not documented in DepthImage.hpp, maybe usecs like DepthData?
+      uint64_t stampUsecs = data->timestamp;
+      depth_image_.header.stamp.fromNSec(stampUsecs * 1000);
       std::copy(
         reinterpret_cast<const uint8_t*>(data->data.data()),
         reinterpret_cast<const uint8_t*>(data->data.data()) + depth_image_.height * depth_image_.step,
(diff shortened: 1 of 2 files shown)


hardware/pico_driver/src/pico_driver.cc
```diff
@@ -430,10 +432,21 @@ class PicoDriverL2 : public PicoDriver, public royale::IExtendedDataListener {
...
+               && data->getIntermediateData() != nullptr) {
+      commonStamp.fromNSec(
+        std::chrono::duration_cast<std::chrono::nanoseconds>(data->getIntermediateData()->timeStamp).count());
+    }
+
     // If we have depth data, use the same mechanism as L1 to push it
     if (data->hasDepthData() && pub_cloud_.getNumSubscribers() > 0
       && data->getDepthData() != nullptr) {
-      cloud_.header.stamp = ros::Time::now();
+      cloud_.header.stamp = commonStamp;
```
marinagmoreira, line 449: why not instead of:
cloud_.header.stamp = commonStamp;

do:
cloud_.header.stamp.fromNSec(
        std::chrono::duration_cast<std::chrono::nanoseconds>(data->getDepthData()->timeStamp).count());

seems that we could compact this instead of creating a new variable

hardware/pico_driver/src/pico_driver.cc
```diff
@@ -443,7 +456,7 @@ class PicoDriverL2 : public PicoDriver, public royale::IExtendedDataListener {
     // If we have a listener and the extended data contains intermediate data, publish it
     if (data->hasIntermediateData() && pub_extended_.getNumSubscribers() > 0
         && data->getIntermediateData() != nullptr) {
-      extended_.header.stamp = ros::Time::now();
+      extended_.header.stamp = commonStamp;
```
marinagmoreira, line 459: same here

marinagmoreira (changes requested):
I'll test this in the lab this week to make sure units are sound :)

hardware/pico_driver/src/pico_driver.cc
trey0: The intent of introducing `commonStamp` was to guarantee that the two outgoing messages that are derived from the same sensor frame get identical timestamps.

I mean, hopefully they would have identical timestamps anyway as assigned by the Pico driver, since the acquisition time was the same, but this way it's guaranteed, and you can rely on the timestamps being identical in any downstream analysis algorithms that want to do something with matched pairs of messages.

marinagmoreira: I would understand if the variable `commonStamp` was used elsewhere in the code but it is not, the contents of `data` don't change.

trey0: My assumptions are that:
- We don't know that `data->getDepthData()->timeStamp` and `data->getIntermediateData()->timeStamp` are the same.
- We don't know in advance which pieces of information (depth, intermediate data) will be defined. It could be either or both.
- In the case that both are defined, we want to guarantee that the two outgoing messages have the same timestamp.

This `commonStamp` solution approach handles it this way:
- If only one piece of information is defined, `commonStamp` gets that timestamp and is used for the one outgoing message.
- If both are defined, `commonStamp` is arbitrarily set to one of them (`data->getDepthData()->timeStamp`) and that is used for both outgoing messages.

I'm not sure I see another way of achieving the design goal that is simpler and equally clear.

marinagmoreira: I was not aware of the design goal "In the case that both are defined, we want to guarantee that the two outgoing messages have the same timestamp" and "If both are defined, commonStamp is arbitrarily set to one of them (data->getDepthData()->timeStamp) and that is used for both outgoing messages."

My assumption was that in the case the timestamps are indeed different, then the desired output timestamp is whatever is more accurate and not necessarily an equal value.

marinagmoreira (approved):
Awaiting lab testing to merge

marinagmoreira:
Screenshot from 2022-03-03 12-59-34
Is this sufficient proof that this works? Stamp matches gnc/ekf grossly and they all match each other

trey0:
Hm, sorry. If you were intending to test whether the DepthImage timestamp is correct per my comment, you would have to enable recording of `/hw/depth_haz/depth_image`, which is not something we normally record, at least for ISAAC surveys. (I think that message is only published if there is a subscriber, so that whole code path may not have been checked much at all.)

However, it's good to know that it's working for the other message types with no surprises!

marinagmoreira:
> Hm, sorry. If you were intending to test whether the DepthImage timestamp is correct per my comment, you would have to enable recording of `/hw/depth_haz/depth_image`, which is not something we normally record, at least for ISAAC surveys. (I think that message is only published if there is a subscriber, so that whole code path may not have been checked much at all.)
> 
> However, it's good to know that it's working for the other message types with no surprises!
Screenshot from 2022-03-07 17-12-33

PR #416, opened 2022-02-06
Added Inverse depth projection factor

rsoussan:
Added inverse depth projection measurement to support the new factor and Jacobians for various projection and backprojection functions.  Added unit tests for each of these along with the factor.  Reorganized some code in vision common for use with the depth projection factor.   More refactoring is being done in a different pr to allow for adding feature id's to the graph for optimization.


bcoltin:
Looks good once build is fixed. Seems like there are multiple tests with the same name.

PR #422, opened 2022-02-17
Added scripts for bayer conversion, histogram equalization, and map creation

rsoussan:
Added scripts to perform bayer to grayscale conversion for a bagfile as well as histogram equalization using normal histogram equalization and CLAHE. Generalized the groundtruth map script to make a map without groundtruth as well.  These all help the analysis of dark data in the ISS.


bcoltin (approved):
Looks good once you fix lint

marinagmoreira (approved):

PR #424, opened 2022-02-22
Adding data share documentation

marinagmoreira:
Didn't elaborate on the documentation because use cases are not clear yet.
Possibly we can add instructions on how to view the data with rqt_bag (http://wiki.ros.org/rqt_bag) -- ROS centric; or instructions on how to convert the data to other formats like .cvs.

### DIFF
--- a/astrobee/readme.md
+++ b/astrobee/readme.md
@@ -1,4 +1,4 @@
-\page astrobee Astrobee
+\page astrobee General Considerations
 
 # Folder description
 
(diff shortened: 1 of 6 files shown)


bcoltin (approved):

trey0:
Thanks, Marina! This looks like a great start. I'm a bit concerned the link to the data will become stale, but we can deal with that when and if it happens. And now that we have this page, we have a place to add more instructions as we start to understand what common use cases people have.

trey0 (approved):

PR #428, opened 2022-02-25
add csv_join.py for joining messages by timestamp

trey0:
Add csv_join.py for joining two CSV files by timestamp, where each CSV file corresponds to a single ROS topic. Where these CSV files could come from is described in the CSV section of [1].

Joining on timestamp is a common analysis task, with the typical example of joining a sensor message with a pose message so you can map the sensor measurements. This script could especially help external users who want to do their own analysis of Astrobee data but might not be ROS experts.

[1] https://github.com/trey0/astrobee/blob/bag_doc/doc/general_documentation/public_data.md

### DIFF
--- a/scripts/postprocessing/csv_join.py
+++ b/scripts/postprocessing/csv_join.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+# Copyright (c) 2017, United States Government, as represented by the
+# Administrator of the National Aeronautics and Space Administration.
+#
+# All rights reserved.
+#
+# The Astrobee platform is licensed under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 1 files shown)


bcoltin:
@rsoussan can you review? How does this fit in with the other bag merging scripts we were discussing monday?

trey0:
@rsoussan could you take a look?

rsoussan (approved):

rsoussan:
> @rsoussan can you review? How does this fit in with the other bag merging scripts we were discussing monday?

Sorry for the delay.  This could go in the new bag_processing package since it's pretty generic, up to you guys.

bcoltin:
Yeah, I think we should put it there. Can you add it? @rsoussan

PR #430, opened 2022-03-01
wait for mapper service validator

marinagmoreira:
This addresses the goal of using the zones for the purpose of defining strict no-fly zones without inflation.


marinagmoreira:
Tested in simulation. For Kibo RPC putposes, the changes in config files are as follows:
Screenshot from 2022-03-01 14-49-55
Tested up to a precision of 0.00002 (was able to move), 0.00001 (keepout warning)

kbrowne15 (approved):

PR #431, opened 2022-03-02
Created bag_processing package

rsoussan:
Added script to convert all bags in a directory from bayer to grayscale.  Made a new bag_processing package and moved some scripts from graph_bag here and added documention to the scripts and the bag_processing readme.


tools/bag_processing/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Package Overview
+The bag processing package provides several helper tools for handling bagfiles.
+
+# Usage Instructions
+For each script, run 'rosrun bag\_processing script\_name.py -h' for further details and 
+usage instructions.
+
+# Scripts
+## apply\_histogram\_equalization\_to\_images
```
trey0, line 11: Single underscores have no special meaning in Markdown and do not need to be escaped like this. (Two or more underscores together is special, boldface or horizontal rule.)

However, as long as we're here, a nice style touch is to use backtick characters around `code_literals` and filenames so they render as monospace.

tools/bag_processing/scripts/convert_all_bayer_bags_to_grayscale.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    parser.add_argument(
+        "-s",
+        "--save-all-topics",
+        dest="save_all_topics",
+        action="store_true",
+        help="Save all topics from input bagfile to output bagfile.",
+    )
+    args = parser.parse_args()
+    for bag in glob.glob("*.bag"):
```
trey0, line 53: This looks like a place where you could use argparse to accept a list of positional arguments so the user could specify an arbitrary glob that their shell would expand, instead of hardcoding a glob.

tools/bag_processing/scripts/convert_bayer_to_grayscale.py
```diff
@@ -30,6 +30,31 @@
...
+    output_bag = rosbag.Bag(output_bag_name, "w")
+
+    with rosbag.Bag(bagfile, "r") as bag:
+        for topic, msg, t in bag.read_messages(topics):
+            if topic == bayer_image_topic:
+                try:
+                    image = bridge.imgmsg_to_cv2(msg, "mono8")
+                except (CvBridgeError) as e:
+                    print(e)
+                gray_image = cv2.cvtColor(image, cv2.COLOR_BAYER_GR2GRAY)
```
trey0, line 49: I would appreciate a comment here, something like:

"We suspect this hard-coded RGB color mapping is not correct for the Astrobee NavCam/DockCam camera Bayer pattern. It may have swapped color channels. A color channel swap has a subtle effect on grayscale output because different color channels have different weighting factors when outputting luminance calibrated for human perception. However, for the purposes of this script, the key requirement is to exactly replicate the onboard debayer conversion performed in the FSW is_camera ROS node, so localization features will be the same regardless of which tool is used to do the conversion."

trey0 (commented):
This looks like a big improvement! I had a few minor comments.

tools/bag_processing/scripts/merge_bags.py
```diff
@@ -77,12 +81,23 @@ def merge_bag(input_bag_prefix, merged_bag, only_loc_topics=False):
 
 
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser()
-    parser.add_argument("input_bag_prefix")
-    parser.add_argument("--merged-bag", default="")
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument(
```
trey0, line 87: For improved flexibility, it seems like you could avoid requiring the bags to have a common prefix if you instead used `argparse` to accept a list of positional arguments using `nargs='+'` [1]. Then the user could easily recover the prefix behavior by specifying `<prefix>*`, which would be expanded by their shell.

(Sorry if this is a dupe. I tried to add a comment like this before, but it looked like it disappeared somehow.)

[1] https://docs.python.org/2.7/library/argparse.html#nargs

rsoussan: I think in practice for us the bags all come with in the same directory, so this prevents having to type each of the prefixes by hand.  (The same comment applies to the convert all bayer images script).  I think for general purpose usage I could add the option to either manually list the prefixes/bags or allow them to be found assuming they are in the same directory, but I might leave that for future work if that's ok as I haven't had that use case yet.

tools/bag_processing/readme.md
rsoussan: Ok, I think when I didn't escape them it didn't render properly as an underscore? I think regardless the backtick solution is better so I'll update each of these with that

tools/bag_processing/scripts/convert_bayer_to_grayscale.py
rsoussan: Done

tools/bag_processing/readme.md
rsoussan: Done

tools/bag_processing/scripts/convert_all_bayer_bags_to_grayscale.py
rsoussan: Done, added glob as behavior if no bags provided to allow default usage to run for all bags in a directory.

tools/bag_processing/scripts/merge_bags.py
rsoussan: Done, added automatically finding prefixes as default behavior if no prefixes are provided

trey0 (approved):
Looks good from here!

marinagmoreira (approved):

PR #435, opened 2022-03-03
changing bayer encoding

marinagmoreira:


trey0 (approved):
Great! Exactly in sync with our conversation earlier.

My only request was to better explain the evidence in the bug report :) Thanks!

PR #436, opened 2022-03-03
public_data.md: add and update several topics

trey0:
One thing it documents is the script in https://github.com/nasa/astrobee/pull/428


trey0:
@bcoltin This may have been waiting on #428 but unblocked now.

@rsoussan If you move `csv_join.py` or other scripts, which is a great idea, we should remember to update the usage example in `public_data.md`.

bcoltin (approved):

PR #437, opened 2022-03-04
Disable mapper

marinagmoreira:


bcoltin:
Hi Marina, what is the reason to not start the timers and subscriptions, instead of not launching the mapper node from the launch file?

marinagmoreira:
> Hi Marina, what is the reason to not start the timers and subscriptions, instead of not launching the mapper node from the launch file?

Because some of the services are needed by the choreographer (like the get collision distance)

bcoltin (approved):

bcoltin:
Got it, thanks. Btw, I think the Initialize is only called on startup, so enabling it through reconfiguring probably won't work currently.

PR #439, opened 2022-03-04
Remove low movement images

rsoussan:
Added script to remove sequential images with low movement. This helps prune redundant images before running the mapping pipeline.


bcoltin:
@rsoussan how does this fit in with select_images? Should we upgrade select_images to do this instead of adding a new binary, to keep the workflow the same?

rsoussan:
> @rsoussan how does this fit in with select_images? Should we upgrade select_images to do this instead of adding a new binary, to keep the workflow the same?

Um, they operate pretty differently and have some different usage options.  I think my plan was to update the documentation once all the changes were added, and to do some more extensive testing then.  Right now this script works pretty well though on the bags I've tested with, but I'd want to do some further full pipeline tests before recommending it over select images

bcoltin:
So the plan is to eventually replace select_images with this, but we are not sure it's ready yet? I'm just worried by adding more options we are just making map building more complicated. I think the ideal case would be if we had one program that did everything you need to get the right images.

rsoussan:
> So the plan is to eventually replace select_images with this, but we are not sure it's ready yet? I'm just worried by adding more options we are just making map building more complicated. I think the ideal case would be if we had one program that did everything you need to get the right images.

Ok, yeah the long term plan would be to replace select_images with this, I think after some further testing and tuning of the optical flow difference parameter we could remove select_images as I don't see a case where it would be preferential to this script. I mostly just don't want to interfere with Andres' current pipeline until I have done some more testing. I could ping him though and see if it is already more useful, in which case we could go ahead and remove select_images

bcoltin:
Ok. I would suggest then, let's keep this as a pull request and wait to merge until it is ready to replace select_images in the pipeline. How does that sound?

rsoussan:
> Ok. I would suggest then, let's keep this as a pull request and wait to merge until it is ready to replace select_images in the pipeline. How does that sound?

Ok that's fine with me

localization/camera/CMakeLists.txt
```diff
@@ -40,19 +32,17 @@ find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system)
 find_package(Eigen3 REQUIRED)
 
+# Find OpenCV
+SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/../../cmake")
```
marinagmoreira, line 36: I'm thinking maybe we should add this to catkin config --cmake-args, such that we never have to add this again..

marinagmoreira:
Is it on purpose that the PR changes multiple CMake files instead of just adding the new tool?
I just tested it and works great! I think the distance closer to what has been done manually is 15cm. It does include more images during rotation, but that might actually be useful to get better map results?

marinagmoreira:
Is it on purpose that the PR changes multiple CMake files instead of just adding the new tool?
I just tested it and works great! I think the distance closer to what has been done manually is 15cm. It does include more images during rotation, but that might actually be useful to get better map results?

rsoussan:
> Is it on purpose that the PR changes multiple CMake files instead of just adding the new tool? I just tested it and works great! I think the distance closer to what has been done manually is 15cm. It does include more images during rotation, but that might actually be useful to get better map results?

I had to clean up some of the CMake files to get the build and tests to work on Ubuntu 20, there was the opencv linking error I think I had mentioned before with different versions of opencv.  Additionally while doing that, some packages when their dependencies were cleaned up were actually missing some dependencies so I added those where needed.

Nice, is the 15 the value passed to the script?  I'm surprised I had it at 0.1 as a default, we should update that if so.

marinagmoreira:
> s to get the build and tests to work on Ubuntu 20, there was the opencv linking error I think I had mentioned before with different versions of opencv. Additionally while doing that, some packages when their dependencies were cleaned up were actually missing some dependencies so I added those where needed.
> 
> Nice, is the 15 the value passed to the script? I'm surprised I had it at 0.1

The value passed in the script I'm assuming it's meters, so 15cm is 0.15m, I wouldn't bother changing since it's not certain that the ideal spacing is 0.15 :)

marinagmoreira (approved):

bcoltin:
We had said before we would replace select_images with this before we merged--- are we ready to do that now? What is the status of this?

rsoussan:
> 

Good point, yes it is ready and I've updated the documentation and removed the select_images tool in favor of remove_low_movement_images.

bcoltin (approved):
Thanks! Looks good when it passes tests.

PR #442, opened 2022-03-07
Fixed deps

rsoussan:
Small pr to fix some dependencies and prevent opencv issues on Ubuntu 18/20 builds if new packages are added to sparse_mapping


bcoltin (approved):
Fix the focal build and then it's good.

marinagmoreira:
Was this tested in the robot? (sometimes you can get runtime errors when missing include packages like nodelet)

rsoussan:
> Was this tested in the robot? (sometimes you can get runtime errors when missing include packages like nodelet)

No not yet, although I think all those dependencies are from when camera may have had a nodelet at some point?  Currently there isn't a nodelet in the package anymore, but I can test anyway on a robot.

Another issue I'm having is that it looks like cv_bridge is bringing in opencv 4 on Ubuntu 20.  This has subtle issues, the code can compile but some tests fail.  I'm guessing simulation works on in Ubuntu 20 though?  I think the only solution is to either upgrade our opencv version or checkout a local copy of cv_bridge and point it to opencv 3 that we use, let me know if you all have any ideas.

bcoltin:
I'd really prefer to avoid either of these, at least until we switch to 20.04. Why do we need this? It was working on all platforms without this PR right?

rsoussan:
> I'd really prefer to avoid either of these, at least until we switch to 20.04. Why do we need this? It was working on all platforms without this PR right?
You can get the build to fail simply by changing the order of some of the dependencies in sparse_mapping.  The linking is pretty fragile right now, and as I'm adding new dependencies builds are breaking.  I'm wondering if there are already issues in the Ubuntu 20 build that we don't see for things that aren't run during simulation or that we don't have tests for.

rsoussan:
(No longer needed since remove_low_movements_images branch  which contained these changes as well was merged to dev)

PR #443, opened 2022-03-08
Add tool to convert bayer to color image

marinagmoreira:


tools/bag_processing/scripts/convert_bayer_bag_to_color.py
```diff
@@ -0,0 +1,92 @@
...
+            if topic == bayer_image_topic:
+                try:
+                    image = bridge.imgmsg_to_cv2(msg, "bayer_grbg8")
+                except (CvBridgeError) as e:
+                    print(e)
+                color_image = cv2.cvtColor(image, cv2.COLOR_BAYER_GB2BGR)
+                color_image_msg = bridge.cv2_to_imgmsg(color_image, encoding="bgr8")
+                color_image_msg.header = msg.header
+                output_bag.write(color_image_topic, color_image_msg, t)
+            # Also save the bayer topic unaltered
```
marinagmoreira, line 53: Hi @rsoussan ! In the bag converter for the greyscale, there is an `else` here which means that the original bayer message isn't saved. Just wondering if it is by design given that if the greyscale script is executed first, color image extraction is not possible.

rsoussan: Oh interesting, yeah I didn't think about the use case at that time.  Maybe we should add an option to save the original image as well, so users can decide if they want smaller bags or larger bags with both images

rsoussan: We could also merge the two scripts here, and pass an option to do grayscale or color conversion

marinagmoreira: Just updated it with the changes :)

tools/bag_processing/scripts/convert_bayer_bag.py
```diff
@@ -74,21 +97,36 @@ def convert_bayer_to_grayscale(
         default="/mgt/img_sampler/nav_cam/image_record",
         help="Output gray image topic.",
     )
+    parser.add_argument(
+        "-c",
+        "--color-image-topic",
+        default="/hw/cam_nav/image_color",
+        help="Output color image topic.",
+    )
```
rsoussan, line 105: It might be a cleaner interface to add the image topic and whether it needs to be converted as separate options.  For the latter, you can use a bool switch:
`    parser.add_argument(
        "--save-color-image",
        dest="save_color_images",
        action="store_true",
        help="Save color images to bagfile.",
    )`

If the action is store_true, it default stores false and vice versa.

tools/bag_processing/scripts/convert_bayer_bag_to_color.py
rsoussan: Looks good! Just one small comment regarding the arg parse options

tools/bag_processing/scripts/convert_bayer_bag.py
marinagmoreira: What I did now was if we set an empty topic, to ignore it, so:
````
rosrun bag_processing convert_bayer_bag.py 2022-02-28-18-46-30.bag -s -k -c ""
````
does not record color image

Comment (rsoussan) [reply tools/bag_processing/scripts/convert_bayer_bag.py:105]: Right, I was thinking to avoid the empty parenthesis, as the bool switch is probably a more common usage pattern. The difficulty though is coming up with different abbreviations for saving a color/grayscale image vs. setting the color/grayscale image topic.  My guess is the topics won't be changed as often, so reserving the -c/g abbreviation for saving the color/grayscale images and not having an abbreviation for the topics might be ok

**Comment by trey0** [comment]
Some other nice-to-haves (might not make the cut to bother with):
- Handle both NavCam and DockCam topics in a single pass
- Script name is no longer quite right: `debayer_bag.py`?

**Comment by trey0** [comment]
> * Script name is no longer quite right: `debayer_bag.py`?

(Never mind, I see it was already changed!)

File: tools/bag_processing/scripts/convert_bayer_bag.py (hunk 2 above)
Comment (marinagmoreira) [reply tools/bag_processing/scripts/convert_bayer_bag.py:105]: yeah, that's actually the reason I didn't add separate parameters for it, I started doing it and couldn't come up with names lol. Maybe use caps, like disable is -C and -G?

Comment (rsoussan) [reply tools/bag_processing/scripts/convert_bayer_bag.py:105]: Oh, that's already taken care of with the bool switch - 
`script -s` 
enables s and 
`script` 
runs without s.  We can add a comment in the parameter description to make that more clear, like 
`Default does not perform color conversion`
I'm actually curious now if arg parse is smart enough to say something like that, our text setting displays default values.

Comment (marinagmoreira) [reply tools/bag_processing/scripts/convert_bayer_bag.py:124]: got it, I didn't read your comment properly earlier xD

Comment (trey0) [reply tools/bag_processing/scripts/convert_bayer_bag.py:134]: Hi, Marina- I like having these bool flags but suggest they should be called "--no-color" and "--no-gray" instead of "-c" and "-g". For whatever reason, I have a strong prior that "-c" should *enable* color so it hurts my brain to have it negated like that.

**Comment by rsoussan** [comment]
Lgtm, might want to consider Trey's suggestion for c/g vs --no-color and --no-gray

**Review by rsoussan — APPROVED** [review]
Lgtm, might want to consider Trey's suggestion for c/g vs --no-color and --no-gray. I'll approve and let you all decide

File: tools/bag_processing/scripts/convert_bayer_bag.py [hunk 3]
```diff
@@ -0,0 +1,168 @@
...
+        help="Bayer image topic name.",
+    )
+    parser.add_argument(
+        "--no-gray",
+        dest="disable_gray",
+        action="store_true",
+        help="Disable grayscale conversion.",
+    )
+    parser.add_argument(
+        "-g" "--gray-image-topic",
```
trey0, line 120: Missing comma typo?

tools/bag_processing/scripts/convert_bayer_bag.py
```diff
@@ -0,0 +1,168 @@
...
+
+import cv2
+import rosbag
+import rospy
+import utilities
+from cv_bridge import CvBridge, CvBridgeError
+from sensor_msgs.msg import Image
+
+
+# TODO(rsoussan): Update/remove comment when color conversion issue resolved.
```
trey0, line 34: I'm noticing this comment is specifically about grayscale conversion, so it would make more sense if moved next to the line with `cv2.COLOR_BAYER_GR2GRAY`, and since we are more certain about the color channels, we could also clarify that `cv2.COLOR_BAYER_GB2GRAY` would be correct except for backward compatibility.

trey0 (approved):
This looks great, just had a couple of tiny comments. Thanks!

tools/bag_processing/scripts/convert_bayer_bag.py
marinagmoreira: oh weird, it might have been the auto linter

Issue #429, opened 2022-02-28
Bags for public release should pass "rosbag check"


trey0:
When we release a bag, we should do our best to minimize pain points for external users trying to analyze the data. Here are some known issues to address:

- [x] We should fix missing message definitions per [1]. Marina found a fixer script with usage described at [2]. However, it currently doesn't work right on bags that contain obsolete message definitions.
- [x] We should consider proactively migrating the messages in the bag to the latest message definitions, so they can be analyzed using the latest FSW build. (Or figure out a better approach.)

As a reminder, when these problems exist in the bag you can easily get the rosbag tools crashing with a confusing bogus error message, even if you are only trying to process a different message that doesn't have the problem.

[1] https://github.com/nasa/astrobee/issues/402
[2] https://github.com/nasa/astrobee/blob/develop/doc/general_documentation/public_data.md

marinagmoreira:
I'm not sure what you mean with "doesn't work right on bags that contain obsolete message definitions", it works for me using the command:

````
python fix_bag_msg_def.py --use-local-defs /home/mgouveia/data/bags/original/20210326_1653_phase1Loc_survey_bay_6_attempt_2.bag out.bag
````

it does generate warnings:

````
Writing out fixed bag ..
 out.bag                                                                                                                                                                                                 0%   0.0 KB /  32.0 GB --:-- ETAWARNING: For type [ff_msgs/FlightMode] stored md5sum [21af9f940eeb9407d4d58829e8cfc200] does not match message definition [0bb389101a5f30087bd644e6596d8e8e].
  Try: "rosrun rosbag fix_msg_defs.py old_bag new_bag."
WARNING: md5sum of loaded type [ff_msgs/FlightMode] does not match that specified
 out.bag                                                                                                                                                                                                 3%   1.1 GB /  32.0 GB 03:25 ETAWARNING: For type [isaac_msgs/InspectionActionGoal] stored md5sum [d36308676d466059465a3cf2963a88fe] does not match message definition [b28a9e00b25932317b0c4fe9c202f1c2].
  Try: "rosrun rosbag fix_msg_defs.py old_bag new_bag."
WARNING: md5sum of loaded type [isaac_msgs/InspectionActionGoal] does not match that specified
[...]

PR #447, opened 2022-03-10
Enable backward compatibility with archived telemetry bags

trey0:
I recommend starting review with the [design document](https://github.com/trey0/astrobee/blob/bag_migrate/doc/general_documentation/maintaining_telemetry.md).

I think the scripts involved might be better organized into Ryan's new `bag_processing` package, but I would rather address moving them in a second pass (there are some cross-references that should not be broken).


trey0:
Fixes #429

rsoussan (approved):
Looks good, I agree bag_processing would be a good future home for these.  We should consider running this automatically after activities (at least the fixes for the java msgs) when bags are saved to our servers.  The bags from our recent activity have the cam_sci msg mentioned here for example

marinagmoreira (approved):

PR #449, opened 2022-03-15
organize docs

trey0:
I noticed the gh-pages documentation sidebar index was getting untidy and tried to clean it up a bit. Unfortunately, I don't know how to test what the new index will look like before the PR is merged, but hopefully the intent is clear in each case.


marinagmoreira (approved):
I compiled the doxygen pages locally and it looks great!

PR #457, opened 2022-03-22
Update the name of the namespaced body frame for relative motion

ana-GT:
This PR fixes an error I was getting when trying to use the teleop tool for commanding relative motions for a namespaced Astrobee. Before the PR, if I send a command like this:
` rosrun executive teleop_tool -ns "bumble" -move -relative -pos "0.5 1 0.5"
`
I would get an error like this: 
`(ros.executive./bumble/executive) Executive: Command failed with message: Move goal failed with response: Invalid reference frame`

With this small fix, the error no longer occurs and Astrobee moves as expected when using the relative-pos option.


kbrowne15 (approved):
Thanks for fixing this.

PR #458, opened 2022-03-23
simplify debugging docker image unit tests

trey0:


scripts/docker/test_astrobee.Dockerfile
```diff
@@ -6,7 +6,9 @@ ARG REMOTE=astrobee
 FROM ${REMOTE}/astrobee:latest-ubuntu${UBUNTU_VERSION}
 
 # Run tests
-RUN cd /src/astrobee && catkin build --make-args tests \
-	&& catkin build --make-args test -j1\
+RUN cd /src/astrobee \
+	&& catkin build --no-status --force-color --make-args tests \
+	&& { catkin build --no-status --force-color --make-args test -j1 || true; } \
```
marinagmoreira, line 11: why are we running the tests twice with the errors suppressed the first time?

`catkin build --no-status --force-color --make-args test -j1`
already runs the tests

marinagmoreira:
One should be careful doing this for any local validation purposes given that integration tests will not run.
Docker tests were made for github validation which can't run gazebo (when configuring the build in docker, integration tests are disabled).

marinagmoreira:
[scripts/docker/astrobee_base.Dockerfile](https://github.com/nasa/astrobee/pull/458/files#diff-402a94a1a6a759ca93b31c934cd88a13c01be4cb99877549a95848e047f581d9) and [scripts/docker/astrobee_quick.Dockerfile](https://github.com/nasa/astrobee/pull/458/files#diff-9ac69f3db9ebb271a806387d3add7dc98b818d161f5950bdda7f7162fcd05470) appear changed, but no editing

trey0:
> [scripts/docker/astrobee_base.Dockerfile](https://github.com/nasa/astrobee/pull/458/files#diff-402a94a1a6a759ca93b31c934cd88a13c01be4cb99877549a95848e047f581d9) and [scripts/docker/astrobee_quick.Dockerfile](https://github.com/nasa/astrobee/pull/458/files#diff-9ac69f3db9ebb271a806387d3add7dc98b818d161f5950bdda7f7162fcd05470) appear changed, but no editing

I just cleared the executable bit in the file permissions. I assume it was set accidentally.

marinagmoreira (changes requested):

trey0:
> One should be careful doing this for any local validation purposes given that integration tests will not run. Docker tests were made for github validation which can't run gazebo (when configuring the build in docker, integration tests are disabled).

If there are any specific extra checks you would recommend, I'd appreciate that. Unfortunately, I'm testing on remote hosts and don't currently have a development environment that can run gazebo.

scripts/docker/run_headless.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
marinagmoreira, line 1: Could we make this part of run.sh? Like add an option --headless

scripts/docker/run_tests.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
marinagmoreira, line 1: Given that this script is not exclusive to docker images, I would put it in /scripts, to be used by local setups alike

scripts/docker/test_astrobee.Dockerfile
marinagmoreira: It looks like exactly the same, but we need to add the -j1 argument to make sure only one test runs at a time (if there are multiple tests running simultaneously sometimes rosnodes interfere with each other, or multiple gazebo servers try to start)

marinagmoreira: I'm going to run both commands on my local machine and make sure both have the same number of tests :)

trey0: Ok. The other issue is that, as I understand it, the call in L11 builds formal test result output files that are then collected and evaluated for overall success/failure check in L14. The question I didn't figure out is whether the call in L13 also builds the test result output files in the same way.

scripts/docker/run_headless.sh
trey0: There are other differences between the two scripts.

I wrote them separately this way because I was trying to avoid breaking the existing `run.sh` in case there are scripts that count on its exact current behavior.

It would not be hard to create a combo script that can do the jobs of both `run.sh` and `run_headless.sh`, but I would ideally want to have a free hand to choose the argument semantics that make the most sense across the wider range of use cases, without requiring backward compatibility with the current `run.sh` arguments. Do you think writing a combo script like that would be worthwhile?

scripts/docker/run_tests.sh
trey0: Done 2dfdbc1e1bb3c8983b5f873aff70104796b7e591

scripts/docker/test_astrobee.Dockerfile
trey0: Another point is that `-j1` might not be needed for testing within Docker if the problem with multiple concurrent tests only arises for integrated tests that are disabled in Docker anyway. Forcing `-j1` could certainly be slowing down the testing, depending on the resources available to the Docker container.

marinagmoreira: We've had this problem with github CI too

scripts/docker/run_headless.sh
marinagmoreira: In my opinion it's worth it to go through the trouble of merging it maintaining backwards compatibility. We kept in the past multiple run scripts in the isaac docker folder and it became hard to maintain when changes were introduced (we since then merged them all under one run.sh). I'll leave it up though :)

scripts/docker/test_astrobee.Dockerfile
marinagmoreira: > Ok. The other issue is that, as I understand it, the call in L11 builds formal test result output files that are then collected and evaluated for overall success/failure check in L14. The question I didn't figure out is whether the call in L13 also builds the test result output files in the same way.

Testing this sort of things is why docker is so useful :)

scripts/docker/run_headless.sh
bcoltin: I agree, I think it would be worthwhile to merge the two scripts and save us headache in the future.

scripts/docker/test_astrobee.Dockerfile
marinagmoreira: So it seems that the L13 command on it's own doesn't actually run the tests.
But using both will just take double the time because the tests can run after the L11 cmd. I don't think we want all our CI pipelines to take double the time. If getting each test output is necessary, I would put it with a | (or) command such that it only happens in case of failure

scripts/docker/run_headless.sh
trey0: Ok, merged `run_headless.sh` features into `run.sh` in e4cf0fd96f757edc8282cff45a8f719a872e4311

The intent is explained at [1]. It changes the argument semantics of `run.sh` slightly for the original use case. (You need to specify `--sim` to run the simulator, and `--sim-args` instead of `--args`.) Otherwise, it just adds useful features.

[1] https://github.com/trey0/astrobee/blob/docker_testing/scripts/docker/readme.md#run-commands-in-the-container

scripts/docker/test_astrobee.Dockerfile
trey0: Should be done in 28c03a22c0b54d0b61fdd411737cfce29a5a6290

The strategy is explained in `run_tests.sh`. I verified it has the desired behavior both when all tests succeed (don't re-run verbosely) and when a test fails (do re-run verbosely). Hopefully the CI testing will behave the same way.

Note: I think actually running the tests (versus compiling FSW and also compiling the tests) may be a pretty small proportion of the overall CI process, so running them twice might not affect the total CI time very much. But it's still good to take the easy performance improvement.

trey0:
In order to better test the changes to `test_astrobee.Dockerfile` without dealing with GitHub PRs, I pushed another commit with some changes to `build.sh`. (The changes add more options, but should not change the old behavior.)

If you think those look ok, I'd say it's ready to merge.

scripts/docker/run.sh
```diff
@@ -17,73 +17,190 @@
...
+                                   ;;
+        --focal )                  os="focal"
+                                   ;;
+        --remote )                 tagrepo="ghcr.io/nasa"
+                                   ;;
+        --no-display )             display="false"
+                                   ;;
+        --mount-directory )        mount="true"
+                                   ;;
+        --sim )                    sim="true"
```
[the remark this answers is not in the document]
marinagmoreira: @trey0

marinagmoreira (approved):
Looks good! I'll leave the remaining feedback to your criteria

scripts/docker/run.sh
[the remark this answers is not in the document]
trey0: Done in d3bb3cd4ba0f55d478acb0dd19f0d08aee324930

(I realized a simpler semantics that doesn't require `--sim` or `--no-sim`. Make the sim the default, but the user can easily specify `bash` if they want a shell. Probably should have done it that way from the start.)

PR #460, opened 2022-03-23
Added splice bag tool

rsoussan:
Added tool to splice a bagfile at user selected timestamps.  The tool iterates through the images in a bagfile and allows the user to select splice timestamps and apply the splice points to generate the spliced bags.  This is particularly useful for mapping to splice a bagfile into multiple smaller bagfiles in a more user friendly way.

@amoravargas


bcoltin (approved):
Nice!

marinagmoreira (approved):

Issue #423, opened 2022-02-22
Simulation Drag Model not Receiving Correct Link Velocity


albee:
The drag plugin, `simulation/src/gazebo_model_plugin_drag/gazebo_model_plugin_drag.cc`, is receiving incorrect inertial frame velocity values for the body link. It seems to always get a unit vector, resulting in large computed drag values equivalent to Astrobee zooming around at 1 m/s all the time, ~0.05 N, (~1/3 of typical per-axis thruster input, for reference). Normal drag values should be about 3 orders of magnitude lower for typical Astrobee speeds, so pretty negligible. We initially noticed this when using a mass parameter estimator that was having trouble converging and always estimating high in simulation---when removing the drag plugin altogether it converges perfectly.

I've gotten this to happen locally for both my FSW 15.1 workspace and a clean `develop` workspace pulled today. I'm still waiting for someone else on our team to confirm this isn't just a weird local problem I'm having.

To pinpoint the issue, line 82 of that file above is giving bad velocity values for `GetRelativeLinearVel()`. Here are some typical values I'm seeing in sim:

![image](https://user-images.githubusercontent.com/5566617/155045316-01537b28-5fb0-408d-9bec-def22699a66d.png)

I can take a look at this deeper it we confirm it's not just me seeing the issue. (Also note that the drag bounces around between negative and positive when Astrobee's actual velocity is low, which is why it's not running off from a constant drag force being applied.)

albee:
@Pedro-Roque In case he gets the chance to check it out.

Pedro-Roque:
Thanks @albee for finding this out! I'll check it on my system by the end of the week too

trey0:
Yeah, thanks @albee, nice catch!

A couple of minor followup points:
- This particular issue could be especially problematic because it accidentally provides a strong artificial stabilizing force for holding position, which could mask controller instability.
- As a non-urgent suggestion, we might want to include a term for air flow. The ISS requirement for cabin air flow is 15-40 feet per minute, and most USOS modules have overhead-to-deck as the nominal flow direction. A very simple improvement to the drag model would be to add a steady flow rate from overhead to deck at 20 feet per minute (7.6 cm/s). The real air flow is of course more complicated and variable, but even a simple term like that would provide qualitatively more realistic behavior when the robot idles propulsion--it won't hold position but instead will systematically drift, usually toward the deck.

Pedro-Roque:
> Yeah, thanks @albee, nice catch!
> 
[...]

PR #462, opened 2022-03-24
Drag model fixed for Astrobee sim

albee:
See #423 

Issue was the `drag_.Normalize()` command also updates the math::Vector3 object itself, so after normalization the magnitude of `drag_.GetLength()` was getting reset to 1. @Pedro-Roque confirmed this was also happening for non-Kinetic builds. Drag forces applied to the model are fixed with this commit.


bcoltin:
@albee Loooks like you might need to import a header for math::Vector3? While you're at it, could you put just the part that is different (getting the vector) inside the #if else and the rest outside it?

marinagmoreira:
> @albee Loooks like you might need to import a header for math::Vector3? While you're at it, could you put just the part that is different (getting the vector) inside the #if else and the rest outside it?

It seems that newer versions of gazebo use ignition::math::Vector3d

albee:
@Pedro-Roque could you check the build for your 20.04 version of the Astrobee sim? Seems like there's a different dependency for Vector3 with newer Gazebo.

Pedro-Roque:
I'll fix it, one sec

Pedro-Roque:
Done - can we re-run the tests?

marinagmoreira:
> Done - can we re-run the tests?

@Pedro-Roque I think it might be a problem on the github side, we got the same problem last week and it fixed itself (but at the time it was flagged in https://www.githubstatus.com/)

marinagmoreira (approved):

marinagmoreira:
I just opened another PR and the actions triggered, so maybe we need to either redo this PR (close and open it again) or do a commit to force the actions to start

albee:
Reopened to retrigger actions (hopefully).

PR #463, opened 2022-03-24
Making mapper start on simulation, allow for enable/disable without restart

marinagmoreira:
Given that we disabled the mapper, the unit test that checked obstacle detection stopped working.
To solve this and also make sure that the mapper keeps working while we have it disabled, I made it such that it is enabled only in simulation mode. I also added a service to allow enable/disable of the mapper (in case we ever need to test it on the robot for a short period - e.g. crew minimal). I also added a small tool to enable configuring a ConfigServer parameter because we didn't have any method to do it (needs work to make it good, but it works for now).


shared/ff_util/tools/config_tool.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <sstream>
+#include <fstream>
+#include <string>
+#include <memory>
+
+// Gflags
+DEFINE_string(ns, "", "Robot namespace");
+DEFINE_string(node, "", "Localization pipeline (no, ml, ar, hr)");
+DEFINE_string(parameter, "nominal", "Flight mode");
+DEFINE_string(type, "", "Path planning algorithm");
```
bcoltin, line 44: I don't think the help messages here are right

marinagmoreira: I actually found out that there is already a reconfigure tool!

bcoltin: Even better!

bcoltin (approved):
Looks good, thanks!

PR #466, opened 2022-03-25
manually trigger some GitHub workflows

trey0:
Two main changes here:
- Add `workflow_dispatch` trigger to the lint and test workflows. It [enables them to be run manually](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). I would like this because I found myself filing bogus PRs in my personal repo just to trigger CI testing, which seems silly. I think Keenan also has done that.
- Add a new workflow whose sole purpose is to be manually triggered if we decide we want to update the base Docker image on `ghcr.io`. Currently the base Docker image is updated only on release, which is a pain when you want to push changes in the code that depend on recent changes in the base Docker image, but of course you can't because then the CI will fail. (But I'm also open to other workarounds for that problem.) A nice side point is I was able to use the latest `build.sh` script to simplify the workflow a lot.

### DIFF
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -3,6 +3,7 @@ name: Build and Test CI for Pull Requests
 on:
   pull_request:
     branches: [ 'master', 'develop' ]
+  workflow_dispatch:
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


marinagmoreira (approved):
If you want to use the docker/build script for gh actions then we should use it on all actions to maintain consistency. (I personally like to see the explicit parameters instead of referring to a 3rd party script, but will leave it to your criteria)

trey0:
> If you want to use the docker/build script for gh actions then we should use it on all actions to maintain consistency. (I personally like to see the explicit parameters instead of referring to a separate script, but will leave it to your criteria)

I agree it's better to be consistent and I have a mild preference to consistently use the `build.sh` script to abbreviate the calls that are currently copy/pasted in many places. ("Don't Repeat Yourself.") Task captured as #467.

PR #468, opened 2022-03-29
unit test for bag migration

trey0:


marinagmoreira (approved):
I made the test use rostest to be able to use the format:
`rostest package_name test_name`

trey0:
> I made the test use rostest to be able to use the format: `rostest package_name test_name`

I set up `nosetest` detection following the [ROS unit testing guide](http://wiki.ros.org/unittest). With that setup, I found I could invoke the test with `(cd tools/bag_processing/test && nosetests -v)` or `./scripts/run_tests.sh bag_processing`. 

I suppose either approach will work equally well with the CI testing scripts.

I'm not sure I fully understand the trade-offs between these different approaches. Is there an advantage to using `rostest` detection?

marinagmoreira:
> > I made the test use rostest to be able to use the format: `rostest package_name test_name`
> 
> I set up `nosetest` detection following the [ROS unit testing guide](http://wiki.ros.org/unittest). With that setup, I found I could invoke the test with `(cd tools/bag_processing/test && nosetests -v)` or `./scripts/run_tests.sh bag_processing`.
> 
> I suppose either approach will work equally well with the CI testing scripts.
> 
> I'm not sure I fully understand the trade-offs between these different approaches. Is there an advantage to using `rostest` detection?

rostest is what we use throughout the repo (http://wiki.ros.org/rostest) - nice to maintain consistency with other nodes. You can also invoke rostest from any path as long as you source the project.

PR #479, opened 2022-04-20
Fixed lua table python script.

kbrowne15:


marinagmoreira (approved):

scripts/build/luaTable.py
```diff
@@ -3,7 +3,7 @@
 Output Python data structure as a Lua table constructor.
 """
 
-from io import StringIO
+from cStringIO import StringIO  # fmt: skip
```
trey0, line 6: This change breaks Python 3 compatibility in order to achieve Python 2 compatibility.

A portable version would be:

```
try:
    from io import StringIO
except ImportError:
    # Python 2 backward compatibility
    from cStringIO import StringIO
```

scripts/build/luaTable.py
```diff
@@ -18,7 +18,7 @@ def dumpStream(out, d, lvl=0):
     def w(s):
         out.write(s)
 
-    if isinstance(d, str):
+    if isinstance(d, basestring): # fmt: skip
```
trey0, line 21: Same idea as before. A portable version would be:

At the top of the file:
```
try:
    # Python 2 backward compatibility
    string_types = (basestring,)
except NameError:
    string_types = (str,)
```

Then write each check like this: `isinstance(d, string_types)`

scripts/build/luaTable.py
kbrowne15: io also exists in python 2. Thus, this doesn't actually fix the original error so I'm unsure if it would fix the original error for python 3 as well. I completely agree with making this python 3 compatible. However, the xgds_planner2 scripts do not work with python 3. Due to this, the python script that imports this script fails when using python 3. I'd rather not make this python 3 compatible until the xgds_planner2 works with python 3 as I can't guarantee the file generated by python 3 will be correct for what the fsw expects. I have added a note to the top of this script explaining as such and have changed the script to exit if python 3 is detected.

trey0: Ok. I believe our general policy is to support both Python 2 and 3, but if it was already broken for Python 3 (at least at the level of the parent script) it could totally make sense to postpone implementing Python 3 support while fixing bugs.

I see what you mean about `io` in Python 2. I did a bit more research. In Python 2, there are three different implementations of a `StringIO` class: `StringIO.StringIO`, `cStringIO.StringIO`, and `io.StringIO`. The first two are ancient and designed to be completely interchangeable at the API level (just different implementations). The last one apparently has a different API, and it's the only one available in Python 3. If you were quickly porting to Python 3 you certainly might think that you could just drop in `io.StringIO` to replace one of the others without realizing the API differences. Possibly some calls are needed like `.encode("utf8")` and `.decode("utf-8")`.

Anyway, I'm onboard to go ahead as-is.

scripts/build/luaTable.py
trey0: Ok.

trey0 (approved):
Looks good!

PR #483, opened 2022-04-28
auto-exposure; remove bayer service and control it only with the config

marinagmoreira:
Auto-exposure algorithm based on https://www.araa.asn.au/acra/acra2007/papers/paper84final.pdf,
which has a reference implementation in https://github.com/rishchou/auto_exposure_ROS/blob/master/src/auto_exposure_control.py

### DIFF
--- a/astrobee/config/cameras.config
+++ b/astrobee/config/cameras.config
@@ -105,9 +105,9 @@ nav_cam_to_sci_cam_timestamp_offset = robot_camera_calibrations.nav_cam_to_sci_c
 
 -- Auto-exposure parameters
 desired_msv = 2.5;
-k_p         = 0.05;
-k_i         = 0.01;
-max_i       = 3;
+k_p         = 20.0;
+k_i         = 2.0;
+max_i       = 50.0;
 
 -----------------------------------------------------
 --  PicoFlexx cameras support the following modes
--- a/astrobee/config/cameras.config
+++ b/astrobee/config/cameras.config
@@ -34,25 +34,27 @@ nav_cam = {
   device               = "/dev/nav_cam",
   gain                 = robot_camera_calibrations.nav_cam.gain,
   exposure             = robot_camera_calibrations.nav_cam.exposure,
+  auto_exposure        = true,
   calibration_gain     = 105,
   calibration_exposure = 30,
-  bayer_enable = false,
+  bayer_enable         = false,
   bayer_throttle_ratio = 3
 };
 
 dock_cam = {
-  width=1280,
-  height=960,
-  undistorted_width=3000,
-  undistorted_height=1800,
-  distortion_coeff=robot_camera_calibrations.dock_cam.distortion_coeff,
-  intrinsic_matrix=robot_camera_calibrations.dock_cam.intrinsic_matrix,
-  device="/dev/dock_cam",
-  gain=            robot_camera_calibrations.dock_cam.gain,
-  exposure=        robot_camera_calibrations.dock_cam.exposure,
+  width                = 1280,
+  height               = 960,
+  undistorted_width    = 3000,
+  undistorted_height   = 1800,
+  distortion_coeff     = robot_camera_calibrations.dock_cam.distortion_coeff,
+  intrinsic_matrix     = robot_camera_calibrations.dock_cam.intrinsic_matrix,
+  device               = "/dev/dock_cam",
+  gain                 = robot_camera_calibrations.dock_cam.gain,
+  exposure             = robot_camera_calibrations.dock_cam.exposure,
+  auto_exposure        = false,
   calibration_gain     = 105,
   calibration_exposure = 30,
-  bayer_enable = false,
+  bayer_enable         = false,
   bayer_throttle_ratio = 3
 };
 
@@ -100,6 +102,12 @@ sci_cam = {
 
 nav_cam_to_haz_cam_timestamp_offset = robot_camera_calibrations.nav_cam_to_haz_cam_timestamp_offset;
 nav_cam_to_sci_cam_timestamp_offset = robot_camera_calibrations.nav_cam_to_sci_cam_timestamp_offset;
+
+-- Auto-exposure parameters
+desired_msv = 2.5;
+k_p         = 0.05;
+k_i         = 0.01;
+max_i       = 3;
 
 -----------------------------------------------------
 --  PicoFlexx cameras support the following modes
--- a/astrobee/config/cameras.config
+++ b/astrobee/config/cameras.config
@@ -107,8 +107,8 @@ nav_cam_to_sci_cam_timestamp_offset = robot_camera_calibrations.nav_cam_to_sci_c
 desired_msv = 2.5;            -- should be 2.5 as histogram size is 5
 k_p         = 20.0;           -- proportional gain
 k_i         = 2.0;            -- integral gain
-tolerance   = 0.2;            -- msv tolerance to minimize exposure reconfigure
 max_i       = 50.0;           -- maximum i
+tolerance   = 0.2;            -- msv tolerance to minimize exposure reconfigure
 
 -----------------------------------------------------
 --  PicoFlexx cameras support the following modes
--- a/astrobee/config/cameras.config
+++ b/astrobee/config/cameras.config
@@ -105,9 +105,9 @@ nav_cam_to_sci_cam_timestamp_offset = robot_camera_calibrations.nav_cam_to_sci_c
 
 -- Auto-exposure parameters
 desired_msv = 2.5;            -- should be 2.5 as histogram size is 5
-k_p         = 20.0;           -- proportional gain
-k_i         = 2.0;            -- integral gain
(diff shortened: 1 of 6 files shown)


hardware/is_camera/src/camera.cc
```diff
@@ -275,26 +274,58 @@ namespace is_camera {
...
+    // Check if bayer configuration changed
+    EnableBayer(bayer_enable);
+
+    // Auto Exposure Parameters
+    bool auto_exposure;
+    if (!camera.GetBool("auto_exposure", &auto_exposure)) {
+      FF_FATAL("Auto Exposure enable not specified.");
+      exit(EXIT_FAILURE);
+    }
+    if (auto_exposure && !auto_exposure_) auto_exposure_timer_.start();
```
bcoltin, line 291: race condition (although unlikely) - should start timer after below parameters initialized

astrobee/config/cameras.config
```diff
@@ -34,25 +34,27 @@ nav_cam = {
   device               = "/dev/nav_cam",
   gain                 = robot_camera_calibrations.nav_cam.gain,
   exposure             = robot_camera_calibrations.nav_cam.exposure,
+  auto_exposure        = true,
```
bcoltin, line 37: Default to false for now!!!

hardware/is_camera/include/is_camera/camera.h
```diff
@@ -47,20 +48,20 @@ struct V4LStruct;
...
   // The size of the image message ring buffer for publishing raw
   // Bayer format images.  The same comments apply about message
-  // lifetime, but in this case we're expecting the subscriber
-  // callback to at most de-Bayer the image before passing it on, so
-  // the buffer doesn't need to be so long.
-  static constexpr size_t kBayerImageMsgBufferLength = 5;
+  // lifetime, but in this case we're publishing at a lower rate and
+  // expecting the subscriber callback to at most de-Bayer the image
+  // before passing it on, so the buffer doesn't need to be so long.
+  static constexpr size_t kBayerImageMsgBufferLength = 10;
```
bcoltin, line 64: is there a reason to expand this one?

hardware/is_camera/src/camera.cc
```diff
@@ -275,26 +278,54 @@ namespace is_camera {
...
+      exit(EXIT_FAILURE);
+    }
 
     if (!camera.GetUInt("bayer_throttle_ratio", &bayer_throttle_ratio_)) {
       FF_FATAL("Bayer throttle ratio not specified.");
       exit(EXIT_FAILURE);
     }
 
-    if (thread_running_) {
+    if (thread_running_ & !auto_exposure_) {
```
bcoltin, line 317: doesn't really matter but I'd use && just by default

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +337,66 @@ namespace is_camera {
...
+    // down the data rate of the camera.
+    if (std::abs(err_p_) > 0.2) {
+      // Calculate PI coefficients
+      err_i_ += err_p_;
+      if (std::abs(err_i_) > max_i_) {
+        err_i_ = (err_i_ > 0) ? max_i_ : - max_i_;
+      }
+      // Update camera exposure parameter
+      camera_auto_exposure_ = std::round(camera_exposure_ + k_p_ * err_p_ + k_i_ * err_i_);
+      if (camera_auto_exposure_ < 0) camera_auto_exposure_ = 0;
```
bcoltin, line 390: I think we should experiment in the lab to find tighter bounds. 0 makes no sense (although it might do something reasonable), 300000 would likely take too long and our framerate would be too slow / blurred (assuming the driver actually used this which is unlikely).  I am worried it will be too dark / bright and the rate will keep drifting, then when lighting conditions return to normal and we try to return it could take a long time to converge back to the right values.

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +337,66 @@ namespace is_camera {
...
+    sensor_msgs::ImagePtr& grey_image = img_msg_buffer_[(img_msg_buffer_idx_ - 1) % kImageMsgBuffer];
+
+    cv::Mat input(kImageHeight, kImageWidth,
+                     cv::DataType<uint8_t>::type,
+                     &(grey_image->data[0]),
+                     kImageWidth);
+
+    // Calculate Histogram
+    cv::Mat hist;
+    int histSize = 5;
```
bcoltin, line 363: I think what you actually want is to have histSize in the config file. Then desired_msv would be histSize / 2. (?)

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +337,66 @@ namespace is_camera {
...
+      mean_sample_value += hist.at<float>(i) * (i+1);
+    }
+    mean_sample_value /= (kImageHeight * kImageWidth);
+
+    // Control Auto exposure with a PI controller
+    err_p_ = desired_msv_ - mean_sample_value;
+
+    // Don't change exposure if we're close enough. Changing too often slows
+    // down the data rate of the camera.
+    if (std::abs(err_p_) > 0.2) {
```
bcoltin, line 382: might also want this in config file

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +337,66 @@ namespace is_camera {
...
+
+      v4l_->SetParameters(camera_gain_, camera_auto_exposure_);
     }
-    bayer_enable_ = req.enable;
 
-    res.success = true;
-    return true;
+    // Publish exposure data
+    std_msgs::Int32 exposure_msg;
+    exposure_msg.data = camera_auto_exposure_;
```
bcoltin, line 398: should add gain like we discussed

bcoltin (changes requested):

hardware/is_camera/src/camera.cc
marinagmoreira: sounds good! We do have a max_i, so it minimizes the recovery time

hardware/is_camera/include/is_camera/camera.h
bcoltin: Yeah I would keep it smaller unless we know a reason to expand it (uses a decent amount of memory)

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +341,67 @@ namespace is_camera {
...
+
+    // Get last generated image from buffer (index incremented after image completed)
+    sensor_msgs::ImagePtr& grey_image = img_msg_buffer_[(img_msg_buffer_idx_ - 1) % kImageMsgBuffer];
+
+    cv::Mat input(kImageHeight, kImageWidth,
+                     cv::DataType<uint8_t>::type,
+                     &(grey_image->data[0]),
+                     kImageWidth);
+
+    // Calculate Histogram
```
trey0, line 365: Arguably we don't need a super-accurate mean sample value. Here is a faster way to calculate the mean that achieves speedup by only sampling a small proportion of pixels (you can tune how much it skips if 50 seems overly aggressive). 

[1] https://github.com/trey0/astrobee/commit/7ec5245b57ca0f2b8dd711669f3cb8750dd4354d

hardware/is_camera/src/camera.cc
```diff
@@ -306,19 +341,67 @@ namespace is_camera {
...
+
+    // Calculate mean sample value
+    double mean_sample_value = 0;
+
+    for (int i = 0; i < hist_size; ++i) {
+      mean_sample_value += hist.at<float>(i) * (i+1);
+    }
+    mean_sample_value /= (kImageHeight * kImageWidth);
+
+    // Control Auto exposure with a PI controller
```
trey0, line 381: I would tentatively recommend replacing the PI controller with the strategy implemented at [1]. This is from an open source driver of the same type of camera we are using (TIS).

The code there is quite complex in terms of having heuristics for trading off exposure time, gain, and iris setting, but if you ignore the existence of multiple exposure parameters and only think about exposure time, it's quite simple:
- If the mean is within 5% of the desired value, don't adjust
- If the mean is further than 5% from the desired value, adjust by 5% toward the desired value.
- (And clip the parameter to the valid range.)

I think by using division instead of subtraction to compare with the desired value, this approach already copes with gamma correction better. Hard coding 5% makes it less tunable than the PI controller, but the flip side is that we can hope the algorithm's presence here means it was thoroughly tested by the vendor and they didn't find much need to tune it.

As we figured out this afternoon, a feature that might be good as a second pass improvement (but not until we have more data for tuning) is to switch the auto exposure algorithm to increase the gain instead of the exposure time once the exposure time reaches a pre-specified maximum value related to motion blur.

[1] https://github.com/TheImagingSource/tiscamera/blob/master/src/algorithms/auto_exposure.cpp

marinagmoreira: It seems that you can reduce their algorithm (second bullet) to:

exposure=exposure*(2/3+1/3(reference/actual))

bcoltin (approved):

PR #484, opened 2022-04-29
Added scripts to build surf maps in parallel

rsoussan:
Helper scripts for map creation, takes a list of bags or uses all the bags in a directory and generates individual surf maps for these (extracts images, removes low movement images, creates surf map using all images as loop closure candidates).  Should be the second/third step in a mapping pipeline (step 0: convert bayer bags if necessary using convert_all_bayer_bags.py script, step 1: splice bags using splice tool, step 2: run this script to create maps for bags, step 3: merge maps using appropriate merge strategy, combine with existing map).

Note: This pr branches off of the remove low images pr https://github.com/nasa/astrobee/pull/439, so that should be merged first

@amoravargas


trey0:
Hi, Ryan- Wow, this is a lot of work! I didn't try to review it all yet. I have a more tangential comment inspired by the detailed steps in your PR description.

It's great to see the thought that you've put into how all of this fits together to make a streamlined mapping process. But I hope we're not in the position that the best description of that process is hidden in a PR :) . If you didn't already, I would encourage you to capture it in a README, even if your thoughts are not fully baked and there isn't a battle-hardened step-by-step guide to follow. That can iterate and grow.

rsoussan:
> Hi, Ryan- ....

Good point, I've added a readme that go over these steps in more detail.

bcoltin:
Sorry for taking so long to look at this. I'm a bit worried about the use of OpenCV 4 and your comments in the code--- do we know everything works on the robot hardware?

bcoltin:
Sorry for taking so long to look at this. I'm a bit worried about the use of OpenCV 4 and your comments in the code--- do we know everything works on the robot hardware?

rsoussan:
> Sorry for taking so long to look at this. I'm a bit worried about the use of OpenCV 4 and your comments in the code--- do we know everything works on the robot hardware?

I'll test on bsharp to make sure the localization side works fine using a recorded bag - do you think that's sufficient?  
I'll probably also move some scripts from localization_analysis to localization_common so I can link against them and move the map building scripts to sparse_mapping, will probably be cleaner.

bcoltin:
Yeah, if we test on the robot I think that is good. But I guess the tests won't run on older ubuntu versions, even though the code works, since they used opencv3 in cv_bridge?

rsoussan:
> Yeah, if we test on the robot I think that is good. But I guess the tests won't run on older ubuntu versions, even though the code works, since they used opencv3 in cv_bridge?

Ok tested with the localization playback tools on the robot and it worked.  The tests should run on all Ubuntu versions, I made a workaround in the previous pr that this branches from to disable affected tests if opencv4 is found

bcoltin:
Yes, what I'm asking is, will this break running things if we don't have opencv 4? Like on 16.04 native?

rsoussan:
> Yes, what I'm asking is, will this break running things if we don't have opencv 4? Like on 16.04 native?

Oh its the opposite - opencv4 introduces errors for certain tests.  So 16.04 is fine, it's later builds where those tests fail.

bcoltin:
Oh I see got it!

bcoltin (approved):

bcoltin:
Waiting to merge for #439

PR #489, opened 2022-05-13
Fixed BRISK descriptor usage in Sparse Map Database

rsoussan:
Currently the sparse map uses BRIEF database descriptors for the DBOW2 image database since no BRISK feature is provided by DBOW2.  However, the DBOW2 BRIEF length is only half as long as our BRISK features, so the latter half of our BRISK descriptors are not being using when building and querying the database. This fix changes the descriptor length to 512 (from 256 for BRIEF) so that the entire BRISK feature is used for the DBOW2 database creation and image queries. This currently only affects localization since the database isn't being used during mapping.

### DIFF
--- a/localization/sparse_mapping/include/sparse_mapping/FBrisk.h
+++ b/localization/sparse_mapping/include/sparse_mapping/FBrisk.h
@@ -1,0 +1,77 @@
+/* Copyright (c) 2017, United States Government, as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ *
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


rsoussan:
This pr would be a clean long term fix, but I'm still testing so it's still a draft.  If I can confirm this issue, I would expect better accuracy for image retrieval during localization since the entire BRISK descriptor length would be used rather than just the first half.  Memory would increase but only for the image database which is already pretty small, the BRISK descriptors in the map are stored in opencv format so this bug doesn't affect that, which is most of the map storage.

PR #491, opened 2022-05-16
View Map Poses

rsoussan:
Added an option to just view map poses.  Also fixed an issue where if the map visualizer is called from a different directory relative to where the map was built, the user can still view just the map poses (before the images were always loaded and on a failure this would prevent viewing the map poses).

@amoravargas


bcoltin (approved):
Nice! This was always annoying.

PR #501, opened 2022-05-27
Bagfix all

marinagmoreira:
* Fix all does entire pipeline
 * Uniform bag_processing package

Still a WIP, needs tons of testing and more flexibility


trey0:
Hi, Marina-

Big picture thought: I designed `rosbag_fix_all.py` with a particular use case in mind, namely that it could be something the Astrobee Facility team could run on raw bag files before they go through the public release process. Since the raw files are often already many GB in size and a big pain to download, I think for that use case it's important not to bloat the "fixed" bag output beyond the size of the raw bag.

Some reprocessing steps in your WIP script would increase the bag file size, so they might be better packaged separately as something the user could run after they download the fixed bag.

Also, the way this script currently works is that it takes an all-topics bag as input and produces an all-topics bag as output (possibly adding one new post-processed topic, or deleting one topic and replacing it with another). But depending on the use case, it can be nicer to output smaller separate bags for each new processed topic that take up less space on disk (not duplicating all the other topics) and can be processed more efficiently by downstream scripts.

In fact, from an ease-of-use and efficiency perspective, the best approach is often to have on-demand processing... like if a script expects RGB images as input, you just tell it the image topic to use, and if the images it receives happen to have a bayer encoding, the script itself can detect this and knows how to apply the necessary debayering, so the user running the script doesn't need to explicitly specify that, and you never need to write a huge bag of 3x-size debayered RGB images to disk. (Of course, automatic on-demand isn't always practical, but it sure can be nice.)

Two other minor comments:
- There is a minor conflict between this PR and #496, which ironically was an attempt to address one of your concerns.
- The Queen calibration changes seem unrelated, might better put them in a separate PR that could go right through.

trey0:
> Please note that this PR is a Draft and not suitable for review. My idea for the final form would address your comments. I still want to make it possible to skip any processing pipeline and have specific profiles for specific needs (e.g. for map building we want to filter unnecessary topics at the end). I am aware of the conflicting PR, I can merge it on this PR no problem

Well, sure, I didn't try to do a complete review, but I thought that the main reason to file a Draft PR is for team awareness and feedback, right?

Maybe I didn't convey properly the spirit of my comments. It's more that I'm not super-satisfied with the way `rosbag_fix_all.py` works already. It's kind of a brittle mess that can be inconvenient to use, so I'm kind of hunting for different ideas on how to structure new bag processing features so they don't have to become part of that mess. But if you're convinced this is the right approach, don't let me stop you.

bcoltin:
Maybe we could add a bit more explanation? Especially I noticed there are these calibration(?) coefficients for the haz cam on different robots. How did we get these, how do we recalibrate? What if we want to do this for perch cam or different robots?

hardware/pico_driver/scripts/pico_split_extended.py
```diff
@@ -57,25 +60,47 @@ def pico_xyz_from_extended(
...
+                xyz = pico.xyz_from_distance(distance, xyz_coeff)
+                pts_msg = pico.make_points_msg(msg.header, xyz)
+                out_bag.write(pts_topic, pts_msg, pts_msg.header.stamp)
+
+                amp_msg = pico.make_amp_msg(msg.header, amplitude)
+                out_bag.write(amp_topic, amp_msg, amp_msg.header.stamp)
+
+                k_logger.info("split %5d extended messages", k_logger.count + 1)
+
+                if keep_extended_topic:
```
trey0, line 81: --save-all-topics should probably imply --keep-extended in order to minimize user surprise.

hardware/pico_driver/scripts/pico_split_extended.py
```diff
@@ -57,25 +60,47 @@ def pico_xyz_from_extended(
...
-            xyz = pico.xyz_from_distance(distance, xyz_coeff)
-            pts_msg = pico.make_points_msg(ext_msg.header, xyz)
-            out_bag.write(pts_topic, pts_msg, pts_msg.header.stamp)
+    out_bag = rosbag.Bag(out_bag_path, "w")
 
-            amp_msg = pico.make_amp_msg(ext_msg.header, amplitude)
-            out_bag.write(amp_topic, amp_msg, amp_msg.header.stamp)
+    with rosbag.Bag(in_bag_path, "r") as bag:
+        # for ext_msg in pico.get_msgs(in_bag_path, ext_topic, fast):
+        for topic, msg, t in bag.read_messages():
```
trey0, line 68: If you know you only want topics that constitute a small subset of the data, it is a major performance improvement to specify the topics in the read_messages() call because it can skip deserializing the other messages. That feature was used in the old version. (Here I guess it would be slightly more complicated because the topic list would only be used when --save-all-topics is not set.)

hardware/pico_driver/scripts/pico_split_extended.py
```diff
@@ -57,25 +60,47 @@ def pico_xyz_from_extended(
...
+                out_bag.write(amp_topic, amp_msg, amp_msg.header.stamp)
+
+                k_logger.info("split %5d extended messages", k_logger.count + 1)
+
+                if keep_extended_topic:
+                    out_bag.write(topic, msg, t)
+
+            elif save_all_topics:
+                out_bag.write(topic, msg, t)
+    out_bag.close()
```
trey0, line 86: It is usually better form to use a context manager for file resources "with" statement. One benefit is that it closes the file in case of an exception.

If you don't like deep indentation when opening two files at the same time, you can use "with a() as b, x() as y:" syntax.

hardware/pico_driver/scripts/pico_split_extended.py
```diff
@@ -98,25 +123,57 @@ def main():
...
+        "--keep-extended",
+        dest="keep_extended_topic",
+        action="store_true",
+        help="Save extended topic alongside converted on the new bagfile",
+    )
+    parser.add_argument(
+        "-n",
+        dest="do_nothing",
+        action="store_true",
+        help="Option to not debayer anything and write output",
```
trey0, line 144: I get the idea but I think this script is not related to debayering so it can be expressed better.

tools/bag_processing/scripts/Makefile.rosbag_fix_all
```diff
@@ -28,42 +28,63 @@
...
 	rm $<  # explicitly delete right away to save disk space
 
-# The fix_all_check targets are .PHONY but too much trouble and not
-# necessary to declare that.
+# Debayer the nav cam
+%.debayer.bag: %.migrate_old.bag
+	${ROSBAG_DEBAYER} ${ROSBAG_DEBAYER_ARGS} $< -o $@
+	rm $<  # explicitly delete right away to save disk space
 
-fix_all_check1-%: %.fix_all_pre_check.bag
```
trey0, line 62: This should not be a blocker to merge, but I just wanted to point out the previous way the rules were written, the two checks were able to run in parallel with each other (they are slow). It could be done that way again if you make the two fix_all_check* targets individually depend on migrate_old and add the fix_all_check* targets as extra dependencies of fix_all.

tools/bag_processing/scripts/Makefile.rosbag_fix_all
```diff
@@ -28,42 +28,63 @@
...
-	rosbag check $<
+# Decode the haz cam into point cloud + image
+%.depth_split.bag: %.debayer.bag
+	${ROSBAG_SPLIT_DEPTH} ${ROSBAG_SPLIT_DEPTH_ARGS} $< -o $@
```
trey0, line 83: It's a bummer that these steps will always be run as a pass-through that copies the large bag file, even when the relevant processing features are disabled. Could your `shutil.copyfile()` call be replaced with an O(1) call to `os.rename()` to avoid the copying overhead? If you don't want the subsequent `rm` command to fail, I think you could use something like `[ -e $< ] && rm $<`.

tools/bag_processing/scripts/rosbag_debayer.py
```diff
@@ -21,18 +21,20 @@
```
trey0, line 21: Doc string is no longer quite right, not just grayscale images.

tools/bag_processing/scripts/rosbag_fix_all.py
```diff
@@ -37,31 +37,90 @@ def dosys(cmd):
...
+    inbag_paths = [p for p in inbag_paths if not ".migrate_old.bag" in p]
+    inbag_paths = [p for p in inbag_paths if not ".debayer.bag" in p]
+    inbag_paths = [p for p in inbag_paths if not ".depth_split.bag" in p]
+    inbag_paths = [p for p in inbag_paths if not ".fix_all.bag" in p]
+
     skip_count = len(inbag_paths_in) - len(inbag_paths)
     if skip_count:
         logging.info(
-            "Not trying to fix %d files that already end in .fix_all.bag", skip_count
+            "Not trying to fix %d files that already starts with fix_all_", skip_count
```
trey0, line 57: I think you reverted this change so the message is now inconsistent?

tools/bag_processing/scripts/rosbag_fix_all.py
```diff
@@ -37,31 +37,90 @@ def dosys(cmd):
...
+
+    # Rosbag decode haz cam
+    if decode_haz is "":
+        rosbag_pico_split_extended_args = "-n"
+    else:
+        output_stream = os.popen("catkin_find --first-only bag_processing resources")
+        coeff_path = (
+            output_stream.read().rstrip() + "/" + decode_haz + "_haz_xyz_coeff.npy"
+        )
+        rosbag_pico_split_extended_args = "-s --in_npy " + coeff_path
```
trey0, line 85: Might be good to assert here that `coeff_path` exists. If it doesn't, it will be much more confusing to track down the problem later.

tools/bag_processing/scripts/rosbag_fix_all.py
```diff
@@ -99,12 +158,37 @@ class CustomFormatter(argparse.ArgumentDefaultsHelpFormatter):
...
+    )
+    parser.add_argument(
+        "--decode-haz",
+        default="",
+        help="decode the extended haz cam topic into points and amplitude_int, argument is robot name",
+        type=str,
+    )
+    parser.add_argument(
+        "--filter",
+        help="filter the bagfile with only some topics",
```
trey0, line 175: It's not obvious what format this argument takes and hard to track down even from the source code. I think I figured out you can give it arbitrary options for `rosbag_filter_topics.py` but not 100% sure (and is there any guidance about quoting?). Probably good to clarify.

tools/bag_processing/scripts/rosbag_topic_filter.py
```diff
@@ -146,8 +154,21 @@ def topic_filter(verbose, inbag_path, accept_patterns, reject_patterns, outbag_p
...
     args = parser.parse_args()
-    topic_filter(args.verbose, args.inbag, args.accept, args.reject, args.outbag)
+
+    for inbag_path in args.inbag:
+        # Check if input bag exists
+        if not os.path.isfile(inbag_path):
+            print(("Bag file " + inbag_path + " does not exist."))
+            sys.exit()
+
+        output_bag_name = args.output.format(inbag=inbag_path)
```
trey0, line 166: I like this way of forming the output bag name! However, if `inbag_path` is not just a bare filename but also specifies the directory, I think this won't do the desired thing. Ideally it could be split using `os.path.dirname()` /  `os.path.basename()` and just the basename part plugged into the output name template.

tools/bag_processing/scripts/rosbag_fix_all.py
```diff
@@ -74,7 +133,7 @@ def rosbag_fix_all(inbag_paths_in, jobs, deserialize=False):
             "If you want to try again, clean first: rm *.fix_all_pre_check.bag"
         )
 
-    return ret
+    return 0
```
trey0, line 136: We should be careful to return non-zero on an error. Otherwise, I think the unit tests might pass when they should fail. You could split this return call into two return calls with different values in the two branches of the conditional above.

trey0 (approved):
Hi, Marina- Thanks for all your work pulling this together! You can see I had lots of comments but they are mostly pretty minor and should not be considered blockers, just trying to keep our code quality tip-top if you have a bit of time to polish.

tools/bag_processing/scripts/rosbag_fix_all.py
```diff
@@ -174,15 +174,15 @@ class CustomFormatter(argparse.ArgumentDefaultsHelpFormatter):
...
         "--decode-haz",
-        default="",
         help="decode the extended haz cam topic into points and amplitude_int, argument is robot name",
+        default="",
         type=str,
     )
     parser.add_argument(
         "--filter",
-        help="filter the bagfile with only some topics",
+        help='filter the bagfile. Use with "" quotes. Example:  --filter "--accepts /hw/imu --regects /loc/ml/features"',
```
trey0, line 183: Should be `--accept` and `--reject`

tools/bag_processing/scripts/rosbag_topic_filter.py
```diff
@@ -135,18 +135,16 @@ def topic_filter(verbose, inbag_path, accept_patterns, reject_patterns, outbag_p
     parser.add_argument(
         "-a",
         "--accept",
-        nargs="?",
+        nargs="*",
```
trey0, line 138: Hm. Your change seems intuitively right to be, but (1) I feel like it was probably working before (because the examples are not arbitrary but represent a way I actually ran it at one time), and (2) I find `argparse` semantics sometimes counter-intuitive.

(But by all means, if you tested it both ways and this fixes a problem, thanks for finding it!)

tools/bag_processing/scripts/rosbag_fix_all.py
trey0: Hm, separate thought... I guess to make it more realistic, it could be `"--accept '/loc/*' --reject '/loc/ml/features'"`. Suggesting the change because, in your prior example, the `--reject` argument would have no effect, so it might tend to confuse people about the filtering semantics.

tools/bag_processing/scripts/rosbag_topic_filter.py
marinagmoreira: yup, tested it before and didn't work, then made the change and worked

tools/bag_processing/scripts/rosbag_fix_all.py
marinagmoreira: thanks for catching that!

PR #502, opened 2022-06-02
Added Service to Reset Map

rsoussan:
Added a service that will reset the map - this reloads the map in the localization_nodelet and restarts the localizer instance there, as well as resetting the graph_localizer so that it initializes with the new map and avoids jumping if the new map is not well aligned to the old one. This should be called when the robot is at standstill since the graph_localizer initializes with zero velocity.  The IMU biases are saved and loaded from the previous localization instance.  To use the service, call: 
rosservice call loc/ml/reset_map /full/path/to/map_name.map

Passing 'default' in place of the map filepath loads the default map used on startup.

Testing using localization bag replay tool, remains to be tested in granite lab.


marinagmoreira:
I can test this in the granite lab :)

bcoltin (approved):

bcoltin:
Great, let's test before merging!

What is the plan for how we will both change the exposure and call this? Will that be a second script? Do we want to make sure the exposure is changed before or after changing the map, or does it not matter if we process a few frames in one map with the wrong exposure?

marinagmoreira:
> Great, let's test before merging!
> 
> What is the plan for how we will both change the exposure and call this? Will that be a second script? Do we want to make sure the exposure is changed before or after changing the map, or does it not matter if we process a few frames in one map with the wrong exposure?

So my current plan is to have an ops repo script that changes exposure (same as used for ISAAC10) + switches map right afterwards. The part where it changes exposure, checks the value from the topic /hw/nav_cam_ctrl to make sure the exposure was changed successfully before continuing.

bcoltin:
Sounds perfect!

marinagmoreira (approved):
Testing procedure in the granite lab:
1) Undock Astrobee - station keeping
2) Switch from granite.map to iss.map
3) Move robot using odometry only
4) Switch to original granite.map while station keeping

I redid 2-4 multiple times. The actual switching works no problems (goal of this PR), but in the granite table, without features it behaves pretty badly and it crashed with the walls twice. Once during the switch from iss.map->granite.map after station keeping. The other time, just doing a small movement in iss.map.

PR #505, opened 2022-06-04
public_data.md: minor fixes

trey0:
- Remind user to `source $ASTROBEE_WS/devel/setup.bash` before using `rosrun`. Leaving that out messed up Jonathan. (I typically add it to my `.bashrc` during install. Should we recommend that to all users in our install instructions? It seems very repetitive to write it in all the usage instructions.)
- Format workarounds for the Markdown parser used by doxygen, which is dumber than the one used by GitHub

### DIFF
--- a/doc/general_documentation/public_data.md
+++ b/doc/general_documentation/public_data.md
@@ -1,8 +1,6 @@
 \page using_telemetry Using Astrobee Robot Telemetry Logs
 
-The ISS Astrobee Facility has established an [Astrobee ISS telemetry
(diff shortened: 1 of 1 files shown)


marinagmoreira:
We can add it to the scripts/configure.sh like we do with ROS/CMake path https://github.com/nasa/astrobee/blob/6d8f094aff49acee4783fb5c8658e49f304047ad/scripts/configure.sh#L317

bcoltin (approved):

PR #507, opened 2022-06-06
Release 0.16.6

marinagmoreira:


marinagmoreira:
I merged it directly using the correct hash as there were commits to develop

Issue #513, opened 2022-06-21
Cannot run using docker in ubuntu 20.04


Lbaron980810:
I cannot run the project correctly using docker, and my environment is ubuntu20.04.

By the way, I'm trying to learn the trajectory generation methods from paper "Smooth trajectory generation on SE(3) for a free flying space robot", which is claimed to be applied for astrobee, but I did not find related code. 

Any help will be appreciated!

Here is how I set up the environment:
```
git clone https://github.com/nasa/astrobee.git $SOURCE_PATH
cd astrobee/
git submodule update --init --depth 1 description/media
su root
./scripts/docker/run.sh --remote
```
Here is what I got in the terminal:

```
root@NRSL:/home/nrsl/astrobee# ./scripts/docker/run.sh --remote
+ cd /home/nrsl/astrobee/scripts/docker
+ docker run -it --rm --name astrobee --volume=/tmp/.X11-unix:/tmp/.X11-unix:rw --volume=/tmp/.docker.xauth:/tmp/.docker.xauth:rw '--env="XAUTHORITY=/tmp/.docker.xauth"' '--env="DISPLAY"' --gpus all ghcr.io/nasa/astrobee:latest-ubuntu20.04 /astrobee_init.sh roslaunch astrobee sim.launch dds:=false robot:=sim_pub
... logging to /root/.ros/log/a7283bd6-f12f-11ec-835a-0242ac110002/roslaunch-73f8dc838e1a-1.log
Checking log directory for disk usage. This may take a while.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

xacro: in-order processing became default in ROS Melodic. You can drop the option.
started roslaunch server http://73f8dc838e1a:39723/

SUMMARY
========

PARAMETERS
 * /dock/robot_description: <?xml version="1....
 * /handrail_21_5/robot_description: <?xml version="1....
 * /handrail_30/robot_description: <?xml version="1....
 * /handrail_41_5/robot_description: <?xml version="1....
 * /handrail_8_5/robot_description: <?xml version="1....
 * /iss/robot_description: <?xml version="1....
 * /robot_description: <?xml version="1....
 * /rosdistro: noetic
 * /rosversion: 1.15.14
 * /simulation_speed: 1
 * /use_sim_time: True

NODES
  /
    access_control (nodelet/nodelet)
    arm (nodelet/nodelet)
    astrobee_state_publisher (robot_state_publisher/robot_state_publisher)
[...]

PR #515, opened 2022-06-29
fixing docker run display commands

marinagmoreira:
Fixing issue https://github.com/nasa/astrobee/issues/513 for bionic
Seems like the display settings were not being parsed properly

This bug was introduced in https://github.com/nasa/astrobee/pull/458


trey0 (approved):
Looks great! (This kind of thing is why I hate bash scripting...)

marinagmoreira:
> Looks great! (This kind of thing is why I hate bash scripting...)

My dream is for Github to be able to run gazebo inside a docker image such that we can run the integration tests on PRs :)

PR #519, opened 2022-07-13
Added camera info publisher

iandouglas96:
Added `camera_info` topic to camera driver.  Takes an additional dependency on `camera` to handle loading/reading camera calibration parameters from config.


bcoltin (approved):

marinagmoreira (approved):

marinagmoreira:
Looks good :) let's wait for the CI to finish before merging

trey0:
I'm not familiar with all the relevant ROS conventions. Would it be better to publish `camera_info` once as a latching message, or is there some non-obvious problem with that?

PR #522, opened 2022-07-20
Add rosjava topic /command to rewrite rules

### LINKED ISSUE #521: rosbag_rewrite_types_rules.json should include all HLP (rosjava) message topics that are in use
Jonathan found a bag from Kibo-RPC Astrobee ISS guest science activity that contained the `/command` topic, which was a bad topic (published by rosjava on the HLP) but not in the rewrite rules.

As explained in [Using Astrobee Robot Telemetry](https://nasa.github.io/astrobee/html/using_telemetry.html), bad topics that aren't rewritten can cause Python scripts processing the bag to crash with the error message

> genmsg.msg_loader.MsgNotFound: Cannot locate message [Header]: unknown package [std_msgs] on search path [{}]

which matches what he saw. And you can figure out what the bad topics are with

> rosrun bag_processing rosbag_detect_bad_topics.py in.bag

Here are my recommended steps to address this issue:

- [x] Add the `/command` topic to the rewrite rules.
- [x] Audit the HLP code to check if there are any other missing bad topics. I originally guessed that all the topics produced by the baseline FSW on the HLP might match the wildcard rule `/gs/*`, which worked for all the ISS bags I tested before, but evidently there may be some exceptions. Does anybody have insight into this?
- [x] Check if we can publicly release this bag so we can include it in the bag_processing regression tests in the open source repo. (It's not critical if we can't, just a good best practice. We have a script that strips down an example bag to only one message on each topic so the bags don't waste too much repo space.)

trey0:
Fixes #521


trey0:
(Oops, should have been against develop.)

PR #528, opened 2022-08-09
queen calibration from ISAAC 6 activity

marinagmoreira:
This calibration was performed using ISAAC 6 calibration data (bartlett:/home/mgouveia/data/bags/20220407_Isaac6/queen_calibration) using the tool camera refiner (in geometry_mapper package).
To verify the results a geometry map was made from an ISAAC6 panorama and the texture was correcly placed on the mesh as well as the mesh matched the expected result. (bartlett:/home/mgouveia/data/bags/20220407_Isaac6/geometry_mapping)


oleg-alexandrov:
Looks good to me. I hope this was validated somehow. 

It looks that the nav_cam intrinsics were not changed, which is how it should be, if this camera was calibrated beforehand, as changing this now would invalidate previous maps created with it. Was nav_cam indeed calibrated before this change?

oleg-alexandrov (approved):
This looks reasonable. I requested a clarification in a separate post.

marinagmoreira:
> Looks good to me. I hope this was validated somehow.
> 
> It looks that the nav_cam intrinsics were not changed, which is how it should be, if this camera was calibrated beforehand, as changing this now would invalidate previous maps created with it. Was nav_cam indeed calibrated before this change?

My awareness was that the nav_cam was calibrated beforehand on the queen checkout activity, @rsoussan, can you confirm this? I don't see it explicitly in https://github.com/nasa/astrobee/commits/master/astrobee/config/robots/queen.config.

rsoussan:
> > Looks good to me. I hope this was validated somehow.
> > It looks that the nav_cam intrinsics were not changed, which is how it should be, if this camera was calibrated beforehand, as changing this now would invalidate previous maps created with it. Was nav_cam indeed calibrated before this change?
> 
> My awareness was that the nav_cam was calibrated beforehand on the queen checkout activity, @rsoussan, can you confirm this? I don't see it explicitly in https://github.com/nasa/astrobee/commits/master/astrobee/config/robots/queen.config.

I can confirm, we ran a queen calibration activity a while ago on the ISS and the changes were pretty negligible compared to the values we already had so we didn't change them.

trey0:
Without necessarily slowing down the process on this PR -- I think that going forward, whenever we update calibration parameters for a robot, we should add a comment about the provenance of the information (when and how the data was collected, what method was used for parameter estimation, where to find the analysis report). This would be very helpful for future users to figure out how much they trust the current calibration and whether it's worth their time to work on updates or improvements.

marinagmoreira:
> Without necessarily slowing down the process on this PR -- I think that going forward, whenever we update calibration parameters for a robot, we should add a comment about the provenance of the information (when and how the data was collected, what method was used for parameter estimation, where to find the analysis report). This would be very helpful for future users to figure out how much they trust the current calibration and whether it's worth their time to work on updates or improvements.

I added some more details to the PR message. The pointed locations have all the mentioned analysis performed (I organized things there that it should stay static to the foreseeable time), no actual report was made though...

trey0:
> I added some more details to the PR message. The pointed locations have all the mentioned analysis performed (I organized things there that it should stay static to the foreseeable time), no actual report was made though...

I was thinking more of a comment in the config file next to the parameters themselves, but anything is better than nothing!

PR #531, opened 2022-08-12
Ctl simulink to c

nlinthacum:
Converting Simulink controller to C++


bcoltin:
@nlinthacum if you have a few minutes, could you put a comment here summarizing the current status? (what's working, what still needs to be fixed)

nlinthacum:
> I also outlined this in the ctl.cc file in a little more detail
Values that are correct:
-traj_error_pos,
-traj_quat
-traj_error_vel
-traj_error_omega
-ctl_status
-pos_err
-pos_err_int

Needs fixing:
-traj_error_att (I suspect there is a math error here since my value and the old value are the same values, just at different times)
-body_force_cmd (Not sure what a reasonable margin of error is for this value, so didn't want to put it as working)

Values that couldn't be tested because they use a value that needs fixing (ie. wrong until the 'need fixing' are fixed):
-body_accel_cmd (dependent on body_force_cmd)
-att_err_mag (this has the same math issue as traj_error_att, I believe)
-body_alpha_cmd (dependent on att_Err_mag)
-body_torque_cmd (dependent on body_alpha_cmd)

Let me know if you have any questions, happy to help!

bcoltin:
Perfect, thank you Nick!

bcoltin:
Moved to #565

PR #537, opened 2022-08-23
Fix ground truth bug + limit probability of tf warning while maintain…

marinagmoreira:
…ing publishing rate


bcoltin (commented):
I am probably missing something, but is the only difference in gazebo versions here that GetWorldPose became WorldPose? If so, why don't we assign the world pose to a temporary object and just put the # ifs only around that so we can make these kinds of mistakes less likely in the future?

marinagmoreira:
> I am probably missing something, but is the only difference in gazebo versions here that GetWorldPose became WorldPose? If so, why don't we assign the world pose to a temporary object and just put the # ifs only around that so we can make these kinds of mistakes less likely in the future?

sounds good, we could add it to astrobee_gazebo.cc, do you want to do it on this PR or open a separate issue?

bcoltin:
Whatever you think is best! If you want to do a separate issue then feel free to go ahead and merge this one.

PR #540, opened 2022-09-20
Groundtruth fixes

iandouglas96:
- Fixed issue where `make_map` was called from `make_groundtruth` with the wrong number of args
- Fixed issue where `basename` was used without defining it


rsoussan:
Thanks for the fix! Looks like this has been broken since some refactoring in this PR:https://github.com/nasa/astrobee/commit/86fc10605617c66c5116a7d2b4bb3a4874d96e8e#diff-ec5f495dff400017404ffd7caca6d8285519e0372b60b68f23bf62b9c5e8674d

rsoussan (approved):

PR #541, opened 2022-09-21
Diff for comparison with upstream

iandouglas96:


trey0:
It's usually better to target PRs against develop.

iandouglas96:
> It's usually better to target PRs against develop.

Yeah, I have no intention of actually merging this, I just wanted a pretty diff against the current master release

trey0:
> > It's usually better to target PRs against develop.
> 
> Yeah, I have no intention of actually merging this, I just wanted a pretty diff against the current master release

Ok sorry if I'm missing something here, just trying to help.

I think the plan for deploying your changes is to merge your branch into develop, then merge develop into master. Right now develop is 19 commits ahead of master. So you probably want to make sure develop is merged into your branch, and then the most helpful nice diff to look at would be between develop and your branch. But feel free to forget all this if you have some other arrangement with @marinagmoreira.

marinagmoreira:
can we close this since the fsw release was done?

PR #548, opened 2022-10-12
make undistort_image more convenient to use during pano stitching

trey0:


oleg-alexandrov (commented):
I will suggest documenting the extra options at https://github.com/nasa/astrobee/blob/develop/localization/camera/readme.md. It would be nice to also add a sample invocation which makes use of these new options, and likely saying a sentence about the motivation, so that in the future a reader could reproduce your use case. 

BTW, the doc for this tool does not seem accessible from the auto-generated documentation at https://nasa.github.io/astrobee/html/index.html. I don't know how to fix that, and doing this is beyond the scope of your work now, but at some point we need to ensure all readme files ar reachable from the table of contents.

marinagmoreira:
> I will suggest documenting the extra options at https://github.com/nasa/astrobee/blob/develop/localization/camera/readme.md. It would be nice to also add a sample invocation which makes use of these new options, and likely saying a sentence about the motivation, so that in the future a reader could reproduce your use case.
> 
> BTW, the doc for this tool does not seem accessible from the auto-generated documentation at https://nasa.github.io/astrobee/html/index.html. I don't know how to fix that, and doing this is beyond the scope of your work now, but at some point we need to ensure all readme files are reachable from the table of contents.

this? https://nasa.github.io/astrobee/html/camera.html

oleg-alexandrov:
Marina, you are right. I found this myself as well. 

This tool's documentation is accessible from https://nasa.github.io/astrobee/html/camera.html. I will now make a very quick edit to that doc page (in the dev branch)  to make it show up prettier in the html.

oleg-alexandrov (commented):
As before, I suggest documenting these new options. If that is done, this can be merged.

trey0:
> As before, I suggest documenting these new options. If that is done, this can be merged.

Thanks for the feedback! I updated the documentation.

I very much appreciate your attention to this, since your mapping tools are necessarily quite complex and therefore are not always easy for others to operate. But as a general point, I think for readme files, we're probably better off putting more tutorial-style information and tips on when to use the options vs. in some cases we are essentially copy-pasting the output of `cmd --help`... with the latter I worry that it is redundant and can easily get out of sync. (Perhaps there is an auto-doc tool that could magically address this.)

oleg-alexandrov:
>  vs. in some cases we are essentially copy-pasting the output of `cmd --help`

We need all of that. Good documentation has examples of usage. It also has a discussion for things where user judgement is needed, where sample invocations are not enough. And we need a full list of options. 

I approve the merge. You are welcome to pull it in.

PR #552, opened 2022-10-20
Documentation gen in docker

marinagmoreira:
the command dictionary doesn't compile well on ubuntu 20, so we'll freeze this action using docker

This should fix the action that's failing

### DIFF
--- a/scripts/docker/documentation.Dockerfile
+++ b/scripts/docker/documentation.Dockerfile
@@ -1,6 +1,15 @@
-# This will set up an Astrobee melodic docker container using the non-NASA install instructions.
-# This image builds on top of the base melodic image building the code.
-# You must set the docker context to be the repository root directory
+# This will set up an Astrobee melodic docker container using the non-NASA
+# install instructions. This image builds on top of the base melodic image
+# building the code. You must set the docker context to be the repository root
+# directory
+#
+# This dockerfile is an alternative to `astrobee.Dockerfile`. It uses previous
+# docker image build artifacts to speed up the catkin build step. If, in the
+# future, you were to use a 'quick' astrobee's image as a build cache, be sure
+# to add `catkin clean --orphans` so you don't bloat the cache image down the
+# line.
+
+# This image generates doxygen documentation for the Astrobee project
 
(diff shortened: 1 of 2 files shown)


scripts/docker/documentation.Dockerfile
```diff
@@ -0,0 +1,54 @@
+# This will set up an Astrobee melodic docker container using the non-NASA install instructions.
```
trey0, line 1: (Not melodic, no ROS)

scripts/docker/documentation.Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# This image builds on top of the base melodic image building the code.
+# You must set the docker context to be the repository root directory
+
+ARG UBUNTU_VERSION=18.04
+FROM ubuntu:${UBUNTU_VERSION}
+
+# Install dependencies
+RUN apt-get update \
+  && apt-get install -y \
+  build-essential \
```
trey0, line 11: Dockerfile style guideline: sort apt-get packages alphabetically (like Python isort)

scripts/docker/documentation.Dockerfile
```diff
@@ -0,0 +1,54 @@
+# This will set up an Astrobee melodic docker container using the non-NASA install instructions.
+# This image builds on top of the base melodic image building the code.
+# You must set the docker context to be the repository root directory
+
+ARG UBUNTU_VERSION=18.04
```
trey0, line 5: Unused, could be simplified away

trey0 (approved):
Looks good to me! Comments very minor.

scripts/docker/documentation.Dockerfile
marinagmoreira: oh interesting, I was fully unaware of this, thank you!!

scripts/docker/documentation.Dockerfile
marinagmoreira: it's used on the line after, we use it on other images for passing in multiple ubuntu versions (that overwrites the ARG that is set by default). I can take it out since for this we only use this version, just thought it was a nice template :)

PR #554, opened 2022-10-26
Fix docker image + fix ubuntu 20 install + simplify launchfile + script bug fixes

marinagmoreira:


scripts/docker/run.sh
```diff
@@ -75,31 +77,38 @@ eval set -- "$opts"
...
 mount="false"
 tagrepo="astrobee"
 dry_run="false"
+image=""
 
 # extract variables from options
 while [ "$1" != "" ]; do
     case $1 in
-        --help )                   usage
+        -h | --help )                   usage
```
trey0, line 89: I think use of `getopt` earlier should have normalized `-h` to become `--help`, so this change isn't necessary. (It should be easy to check.) Same comment applies to the other arguments below.

astrobee/launch/sim.launch
```diff
@@ -28,13 +28,14 @@
   <arg name="nodes" default=""/>                <!-- Launch specific nodes   -->
   <arg name="extra" default=""/>                <!-- Inject additional node  -->
   <arg name="debug" default=""/>                <!-- Debug node group        -->
-  <arg name="physics" default="ode"/>           <!-- Physics engine          -->
+  <!-- Physics engine options: ode (gazebo default), bullet, simbody, dart   -->
+  <arg name="physics" default="dart"/>
```
trey0, line 32: Do we need to change this?

I vaguely recall you found that `dart` worked better than `ode` in some corner case, and your testing was partly motivated because you thought `dart` could aid Rika's research, but then when I talked to Rika she thought it wasn't needed and suggested that she had other modeling tricks to make the dynamics in her scenarios better-conditioned.

I guess the bottom line is this seems like something that could break simulation scenarios for some of our users in unpredictable ways, and I'm not clear who benefits. So maybe it's not a good idea to change right now without more discussion. It's already easily available for users who want to switch, correct?

scripts/docker/astrobee_base.Dockerfile
```diff
@@ -46,7 +44,7 @@ RUN apt-get update \
 COPY ./scripts/setup/debians /setup/astrobee/debians
 
 RUN apt-get update \
-  && /setup/astrobee/debians/build_install_debians.sh \
+  && /bin/bash /setup/astrobee/debians/build_install_debians.sh \
```
trey0, line 47: The way you're switching to `bash` here may be the best, most conservative choice in this case, but I wanted to point out that Docker also allows you to switch from `sh` to `bash` for all `RUN` commands in the Dockerfile using this command:

    SHELL ["/bin/bash", "-c"]

scripts/docker/build.sh
```diff
@@ -33,6 +33,7 @@ Build specified docker image targets. Available targets:
 - astrobee_base
 - astrobee
 - test_astrobee
+- astrobee_quick
```
trey0, line 36: I think it would be helpful to document somewhere what your recommended workflow looks like for using the `astrobee_quick` target.
- Do you just always use it by default and it makes things run faster? Or do you have to build `astrobee` at least once before it works to build `astrobee_quick`?
- Is there ever a downside to always using `astrobee_quick`? (Like: "If you think your build is breaking because the cache is corrupted somehow, here's how to clear it.")

trey0 (approved):
Looks good to me! Only some minor comments.

scripts/docker/run.sh
marinagmoreira: I did check , but it doesn't normalize it, it only checks if it's on the allowed list. I was also very confused by this...

trey0: Bummer. Helps to know for future scripts, though!

PR #555, opened 2022-10-27
Added Script to Remove Rotation-Only Image Sequences

rsoussan:
Removes any rotation-only image sequences. Checks sequential images and removes the subsequent image if it fits a rotation-only movement model. Further prunes results by removing any images bounded within a provided threshold by detected rotations. Optionally saves results to subdirectories where each subdirectory contains a different sequence of non-rotation movement, where each sequence is separated by a set of removed rotation images.
Optional -v view mode shows debugging images, with matched points projected from one image to the next using the rotation model. Removed images are labeled with "removing", and a ratio of rotation model error to full pose model error is shown. Example provided below:
remoe_rotation_removed
@oleg-alexandrov @trey0

### DIFF
--- a/localization/sparse_mapping/CMakeLists.txt
+++ b/localization/sparse_mapping/CMakeLists.txt
@@ -185,7 +185,7 @@ target_link_libraries(remove_low_movement_images
   sparse_mapping ${catkin_LIBRARIES})
 
 ## Declare a C++ executable: remove_rotation_only_images
-add_executable(remove_rotation_only_images tools/remove_rotation_only_images.cc)
+add_executable(remove_rotation_only_images tools/remove_rotation_only_images.cc tools/utilities.cc)
 add_dependencies(remove_rotation_only_images ${catkin_EXPORTED_TARGETS})
 target_link_libraries(remove_rotation_only_images
   sparse_mapping ${catkin_LIBRARIES})
--- a/localization/sparse_mapping/CMakeLists.txt
+++ b/localization/sparse_mapping/CMakeLists.txt
@@ -179,7 +179,7 @@ target_link_libraries(parse_cam
   sparse_mapping gflags glog ${catkin_LIBRARIES})
 
 ## Declare a C++ executable: remove_low_movement_images
-add_executable(remove_low_movement_images tools/remove_low_movement_images.cc)
+add_executable(remove_low_movement_images tools/remove_low_movement_images.cc tools/utilities.cc)
 add_dependencies(remove_low_movement_images ${catkin_EXPORTED_TARGETS})
 target_link_libraries(remove_low_movement_images
   sparse_mapping ${catkin_LIBRARIES})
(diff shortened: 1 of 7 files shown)


oleg-alexandrov:
In my experience all images have at least some translation, it is not easy to do purely rotation, given that the bot body center is not the same as the camera center. 

How is the rotation defined? Zero translation, small translation? Is there a threshold? 

What if the bot stays in place? Is that a rotation? 

A scale-independent way of defining "very small translation" is the convergence angle.  A pair of images with a convergence angle of under 1-3 degrees means the robot camera performed almost a pure rotation or it stayed in place.

rsoussan:
> In my experience all images have at least some translation, it is not easy to do purely rotation, given that the bot body center is not the same as the camera center.
> 
> How is the rotation defined? Zero translation, small translation? Is there a threshold?
> 
> What if the bot stays in place? Is that a rotation?
> 
> A scale-independent way of defining "very small translation" is the convergence angle. A pair of images with a convergence angle of under 1 degree means the robot camera performed almost a pure rotation or it stayed in place.

I use a method similar to that used in OpenSFM, where I find a relative affine3d between image pairs using feature matches (optical flow based in this case) and measure how well the pair fits the rotation-only portion of the found transform. This is computed by backprojecting feature points from the first image to the unit plane, rotating, projected to the second image, and measuring the mean error in image space.  Since there are other error sources due to feature tracks not being perfect, distortion effects, intrinsics not being perfect, I compare this projected mean error to the same computed using the full affine3d. If the ratio is below a user defined threshold, the image is labeled as a rotation. This works pretty well on the bag files I've tested with. Obviously nothing is a pure rotation, but there are several parameters in the script that users can adjust.

oleg-alexandrov:
Sounds good. I am not sure if the rotation threshold option and value should be added to the example usage in the doc, and if there's a meaningful way of interpreting it. Or at least, one could add a brief explanation to the doc where the tool is invoked. 

It is also not clear to me what config_path is and where an example of that can be found.

bcoltin (commented):
In addition to Oleg's changes, could you add an explanation of when and why to run this?

rsoussan:
> Sounds good. I am not sure if the rotation threshold option and value should be added to the example usage in the doc, and if there's a meaningful way of interpreting it. Or at least, one could add a brief explanation to the doc where the tool is invoked.
> 
> It is also not clear to me what config_path is and where an example of that can be found.

Ok sounds good, I'll add some more info on the parameters

rsoussan:
> In addition to Oleg's changes, could you add an explanation of when and why to run this?

Sure will do

rsoussan:
> Sure will do

Ok I've updated the documentation and added some extra info on why the script is useful. I tried to keep most of the documentation in the script itself to follow the pattern of having users use `rosrun package_name script_name -h` to learn about the script and its options, and I mention this in the mapping guides.

bcoltin (approved):
Thanks!

oleg-alexandrov:
Looks good. 

I prefer documentation where everything is spelled out, even at the risk of repetition, rather than having the user run the --help option and sort it out from the text dump. This way one doesn't need to get the code and to build it to understand what it does. But that's a personal preference. As long as things are written or dumped somewhere, including how to set up a config file, it is fine with me. No further comments.

PR #562, opened 2022-11-02
feat: adding ROS2 compilation to CMakeLists and Package XMLs

Pedro-Roque:
Starting to add ROS2 compilation by checking environment variable ROS_VERSION.

Tagging @marinagmoreira and @trey0 , maybe we can try this until we face a big roadblock, and then ditch the older version.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,2 +1,4 @@
-FROM ghcr.io/nasa/astrobee:latest-ubuntu20.04
+FROM ghcr.io/nasa/astrobee:latest-rolling-ubuntu20.04
 ENV DEBIAN_FRONTEND=dialog
+
+RUN mkdir /src && mkdir /src/astrobee
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,6 +8,7 @@
 	"remoteUser": "root",
 	"runArgs": [
 		"--network=host",
+		"--privileged",
 		"--cap-add=SYS_PTRACE",
 		"--security-opt=seccomp:unconfined",
 		"--security-opt=apparmor:unconfined",
@@ -40,5 +41,5 @@
 		"zachflower.uncrustify"
 	],
 	"workspaceMount": "source=${localWorkspaceFolder},target=/src/astrobee/src,type=bind",
-	"workspaceFolder": "/workspace"
+	"workspaceFolder": "/src/astrobee/src"
 }
--- a/.github/workflows/ci_ros2.yml
+++ b/.github/workflows/ci_ros2.yml
@@ -1,0 +1,24 @@
+name: Build ROS2 base images
+
+
+on:
+  push:
+    branches: [ 'ros2' ]
+
+jobs:
+
+  rolling:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Log in to registry
+      run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.repository_owner }} --password-stdin
+
+    - name: Build, test, push docker
+      run: ./scripts/docker/build.sh --remote
+                                     --owner ${{ github.repository_owner }}
+                                     rolling_base rolling push_rolling_base push_rolling
+
--- a/astrobee/package.xml
+++ b/astrobee/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package>
+<package format="3">
   <name>astrobee</name>
   <version>0.0.0</version>
   <description>The astrobee package provides system launch functionality.
@@ -13,9 +13,12 @@
   <maintainer email="astrobee-fsw@nx.arc.nasa.gov">
     Astrobee Flight Software
   </maintainer>
-  <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
-  <run_depend>roscpp</run_depend>
+  <build_depend condition="$ROS_VERSION == 1">catkin</build_depend>
+  <build_depend condition="$ROS_VERSION == 2">ament_cmake</build_depend>
+  <build_depend condition="$ROS_VERSION == 1">roscpp</build_depend>
+  <build_depend condition="$ROS_VERSION == 2">rclcpp</build_depend>
+  <exec_depend condition="$ROS_VERSION == 1">roscpp</exec_depend>
+  <exec_depend condition="$ROS_VERSION == 2">rclcpp</exec_depend>
   <export>
   </export>
 </package>
--- a/scripts/docker/build.sh
+++ b/scripts/docker/build.sh
@@ -76,8 +76,12 @@ os=`cat /etc/os-release | grep -oP "(?<=VERSION_CODENAME=).*"`
 build_astrobee_base="false"
 build_astrobee="false"
 build_test_astrobee="false"
+rolling_base="false"
+rolling="false"
 push_astrobee_base="false"
 push_astrobee="false"
+push_rolling_base="false"
+push_rolling="false"
 
 remote="false"
 owner="nasa"
@@ -130,9 +134,17 @@ while [ "$1" != "" ]; do
                                    ;;
         astrobee_quick )           astrobee_quick="true"
                                    ;;
+        rolling_base )             rolling_base="true"
+                                   ;;
+        rolling )                  rolling="true"
+                                   ;;
         push_astrobee_base )       push_astrobee_base="true"
                                    ;;
         push_astrobee )            push_astrobee="true"
+                                   ;;
+        push_rolling_base )        push_rolling_base="true"
+                                   ;;
+        push_rolling )             push_rolling="true"
                                    ;;
         * )                        echo "unknown target '$1'"
                                    usage
@@ -258,10 +270,26 @@ if [ "$astrobee_quick" = "true" ]; then
     build astrobee_quick "${revision}-" "quick-"
 fi
 
+if [ "$rolling_base" = "true" ]; then
+    build ros2/ros2_rolling_base "${revision}-" "rolling_base-"
+fi
+
+if [ "$rolling" = "true" ]; then
+    build ros2/ros2_rolling_deb "${revision}-" "rolling-"
+fi
+
 if [ "$push_astrobee_base" = "true" ]; then
     push astrobee_base "${revision}-" "base-"
 fi
 
 if [ "$push_astrobee" = "true" ]; then
     push astrobee "${revision}-" ""
+fi
+
+if [ "$push_rolling_base" = "true" ]; then
+    push rolling_base "${revision}-" "rolling_base-"
+fi
+
+if [ "$push_rolling" = "true" ]; then
+    push rolling "${revision}-" "rolling-"
 fi
--- a/scripts/docker/ros2/ros2_rolling_base.Dockerfile
+++ b/scripts/docker/ros2/ros2_rolling_base.Dockerfile
@@ -1,0 +1,81 @@
+FROM ubuntu:focal
+
+ENV ROSDISTRO=rolling
+ENV ROS_PYTHON_VERSION=3
+ENV ROS_VERSION=2
+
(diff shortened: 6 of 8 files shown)


trey0:
Great!

I'm a big fan of trying to keep support for both ROS1 and ROS2 in the `ros2` branch such that, once we agree it's on the right track, it could be merged into `develop` and the ROS2 version would be further matured there. My sense is that approach would best support keeping the behavior of the two versions very close and maximizing shared code as much as possible, so that the ROS2 version automatically benefits from ongoing improvements in the ROS1 version, and it's as easy as possible for guest science users to migrate. (And eventually the ROS1 version would be end-of-lifed, but that's probably a ways off.)

As I see it, there could be a mix of approaches for allowing multiple versions of the code for each package to be stored in the same folder of the same repo branch, with the right approach selected on a case-by-case basis for each file:

- For files like `CMakeLists.txt` and C++ code where conditionals are available, alternate versions could be captured that way. Although hopefully for C++ code most of the code can be shared between versions, with conditional compilation only around a few constructs that differ between ROS versions.
- Some commonly occurring cases that require conditional compilation could be factored out into an simple ROS1/ROS2 abstraction layer if that helps make the code more readable and maintainable.
- For file formats that don't support internal conditionals, we could use a trick like keeping two separate files in the repo with `ros1-` and `ros2-` prefixes, and having the build system choose which one to use (e.g., by automatically generating a symlink to whichever version is correct, or simply specifying the right filename in each conditional branch of `CMakeLists.txt`).

trey0:
A minor suggestion: perhaps we could define the variable USE_ROS1 uniformly across all file types (available as an environment variable, as a CMake variable, as a C++ macro specified to the compiler with `-D`, etc.). That way, the common idiom would be:

#if USE_ROS1 // begin ROS1 version
...
#else // begin ROS2 version
...
#endif // end ROS2 version

Pedro-Roque:
Awesome! A hiccup now is package.xml support for both versions - I'm trying to script it in Python to generate the dependencies automatically. I'll report here as it goes...

Pedro-Roque:
> A minor suggestion: perhaps we could define the variable USE_ROS1 uniformly across all file types (available as an environment variable, as a CMake variable, as a C++ macro specified to the compiler with `-D`, etc.). That way, the common idiom would be:
> 
> #if USE_ROS1 // begin ROS1 version ... #else // begin ROS2 version ... #endif // end ROS2 version

I think this could work as long as we don't add too much entropy to the source files. That will be something to check after

Pedro-Roque:
Here goes an ugly attempt at generating compatible XMLs for each ROS version. 

I'd use ROS1.xml and ROS2.xml , then read them on package.py and write them to package.xml depending on the ROS version. 

I'm not sure of how to select specific XMLs with colcon, but if someone has a good idea there, I'm all up for it!

trey0:
> I'm not sure of how to select specific XMLs with colcon, but if someone has a good idea there, I'm all up for it!

My current sketch is that if you have a workspace `~/ros1_ws` your git checkout could live at `~/ros1_ws/git_src` and there could be a script that creates a source folder in the usual location `~/ros1_ws/src`, where `src` is built as follows:
- The folder structure is the same as `git_src`.
- Most plain files are symlinks to the corresponding file with the same name in `git_src`
- If a plain file in `git_src` has a filename starting with `ros1-`, it is symlinked without the prefix. Example symlink: `src/foo/bar/package.xml` -> `git_src/foo/bar/ros1-package.xml`.
- If a plain file in `git_src` starts with `ros2-`, it is just left out of `src`.

The auto-generation of the `src` folder could be part of the `ros1-configure.sh` script, so it is already done before `catkin` is invoked.

Likewise, the `ros2-configure.sh` script could build `~/ros2_ws/src` in the same way, but selecting the prefix it wants, before `colcon` is invoked. (And if you like, the two workspaces could share a single git checkout by symlinking `~/ros2_ws/git_src` -> `~/ros1_ws/git_src`.)

There would be nothing precious in the `src` folder, so it could be blown away and regenerated at any time, like if a new file is added to `git_src`.

trey0:
> Here goes an ugly attempt at generating compatible XMLs for each ROS version.

I salute the attempt to share code between ROS versions, but I think we should evaluate very carefully whether any specific auto-generated code solution is worth the increment in complexity. This might be a case where manually maintaining two distinct files `ros1-package.xml` and `ros2-package.xml` is the simplest solution. (The `package.xml` changes rarely and should not have terrible consequences if the two ROS versions accidentally get out of sync... it would typically cause a compile-time problem.)

If we decide it's worthwhile to auto-generate the two versions, I would recommend basing the auto-generation on a template language like [Jinja2](https://palletsprojects.com/p/jinja/). Like there could be a source file `generic-package.xml` which is a Jinja template with conditional constructs that gets compiled to either `ros1-package.xml` or `ros2-package.xml`. A snippet might look like:

```XML
<?xml version="1.0"?>
<package>
  <name>astrobee</name>
  <version>0.0.0</version>
  <description>The astrobee package provides system launch functionality.
  </description>
  {% if USE_ROS1 %}
     <buildtool_depend>catkin</buildtool_depend>
  {% else %}
     <buildtool_depend>ament_cmake</buildtool_depend>
  {% endif %}
</package>
```

This results in a longer XML file, but one where the generated result is very transparent without needing to understand a separate Python script. It can also handle cases where there are real differences between the ROS versions (like the package dependencies have to be different).

Pedro-Roque:
That is a better idea, will give it a go.

Pedro-Roque:
> > I'm not sure of how to select specific XMLs with colcon, but if someone has a good idea there, I'm all up for it!
> 
> My current sketch is that if you have a workspace `~/ros1_ws` your git checkout could live at `~/ros1_ws/git_src` and there could be a script that creates a source folder in the usual location `~/ros1_ws/src`, where `src` is built as follows:
> 
> * The folder structure is the same as `git_src`.
> * Most plain files are symlinks to the corresponding file with the same name in `git_src`
> * If a plain file in `git_src` has a filename starting with `ros1-`, it is symlinked without the prefix. Example symlink: `src/foo/bar/package.xml` -> `git_src/foo/bar/ros1-package.xml`.
> * If a plain file in `git_src` starts with `ros2-`, it is just left out of `src`.
> 
> The auto-generation of the `src` folder could be part of the `ros1-configure.sh` script, so it is already done before `catkin` is invoked.
> 
> Likewise, the `ros2-configure.sh` script could build `~/ros2_ws/src` in the same way, but selecting the prefix it wants, before `colcon` is invoked. (And if you like, the two workspaces could share a single git checkout by symlinking `~/ros2_ws/git_src` -> `~/ros1_ws/git_src`.)
> 
> There would be nothing precious in the `src` folder, so it could be blown away and regenerated at any time, like if a new file is added to `git_src`.

I don't like the idea of having two srcs symlinked, but in PX4 we have Jinja to autogenerate files from templates. So I'll try this one out.

Pedro-Roque:
@trey0 something like this? Now when the `configure.sh` runs, it generates the corresponding package.xml for the desired ROS version (right now I'm putting ROS1 hardcoded - but I can read an env variable to set it to the desired ROS version). 

Then, instead of keeping package.xml files, we keep the package.xml.jinja (jinja templates), which contain the dependencies for both ROS builds.

Thoughts?

trey0:
Yeah, now that I think about it, I think we could push this templating approach a very long way. It could even trivially subsume auto-selecting between alternate versions of a complete file, like the `file.txt.jinja` template that generates `file.txt` could be written this way:

```
{% if USE_ROS1 %}
  {% include ros1-file.txt %}
{% else %}
  {% include ros2-file.txt %}
{% endif %}
```

However, I had some remaining motivations for suggesting the symlink `src` vs. `git_src`, and I'm not sure what alternative ideas you would offer to address them:
- Auto-generated (or auto-selected) files would be written to `src` rather than `git_src`, so they would be invisible to git as they should be (without needing to add lots of entries to `.gitignore`). The separation would also make it harder to accidentally edit an auto-generated file instead of its template.
- Because the `git_src` source checkout would not be polluted with version-dependent auto-generated files, it could easily be reused across both ROS1 and ROS2 workspaces, so if you make a change, you could easily test the modified code in both workspaces before pushing your changes.

Given those benefits, my inclination would be to give `src` vs. `git_src` a try and see if there are any hidden drawbacks that outweigh those benefits. I think the script to do the symlinking would be quite easy to write. (An example hidden drawback could be confusing your IDE so that it has trouble using an error message to load the offending line of source code in its editor.)

scripts/configure.sh
```diff
@@ -305,6 +305,10 @@ if [[ $native_build == 1 && $armhf_build == 1 ]] ; then
 fi
 
 if [ $native_build == 1 ] ; then
+    echo "Generating XML files with Jinja2..."
+    python3 src/scripts/build/package.py
```
trey0, line 309: Here's a point where Astrobee multi-platform support makes things more complicated. The current Astrobee FSW policy is:
- Python scripts should be written so they are portable to either Python 2.7 (Ubuntu 16.04, 18.04) or Python 3.8 (Ubuntu 20.04). Of course, this means they sometimes have to prioritize maintaining Python 2.7 compatibility over using Python 3.x best practices.
- Executable scripts should start with `#!/usr/bin/env python`. Under Ubuntu, you conventionally run `python` for Python 2.x and `python3` for Python 3.x, but when we run under Ubuntu 20.04, since it doesn't have any Python 2.x interpreter, we install a symlink `python` -> `python3` to make this work in all cases.
- Note that any Python scripts used only in the ROS2 version can safely assume they are running under Ubuntu 20.04+. They don't need to maintain Python 2.x compatibility, and should explicitly call `python3`.

Pedro-Roque: Well caught! I agree, This should be executables or, if ROS2, calling python3 directly. Will change that.

trey0:
> Amazing inputs @trey0 ! Thanks a lot! So I'm thinking on doing something just to try the two approaches. I'll create a ROS2-develop branch which is blank (no packages there), and then try these scripts, try having symlinking, try without, and then we see what we could do there. I'll open a separate PR for that I'll tag here. Otherwise this PR will quickly go out of context.

Last night I started working on [an implementation of the symlinking script, available here](https://github.com/trey0/astrobee/commit/c4cafd764ddb9d80d4d5d52bb5cf6eacbb4f2d85), sorry if we are doing redundant work.

Maybe it's better to have most of the conversation tied to an issue rather than a PR? I haven't done this much before.

trey0:
> I'll create a ROS2-develop branch which is blank (no packages there), and then try these scripts, try having symlinking, try without, and then we see what we could do there.

Some quick notes:
- I tested [the draft symlinking script](https://github.com/trey0/astrobee/commit/c4cafd764ddb9d80d4d5d52bb5cf6eacbb4f2d85) and it seemed like  `catkin` was happy to build FSW from the resulting `src` folder full of symlinks. But I'm sure this could be more thoroughly tested to look for any weird corner cases.
- Since ROS2 package coverage will be minimal at first but we don't want to blow away all of the ROS1 packages pending their ROS2 support, I was thinking we could get the ROS2 version jump-started by renaming all of the `package.xml` files to `ros1-package.xml`. Then they won't show up in the ROS2 version of the `src` folder, and my hope is that would cause `colcon` to ignore them. (So it would start as "successfully" building no packages.)
- After that, I guess the ROS2 version of `configure.sh` would need to be reworked to invoke ROS2 build tools, and we could start migrating packages. Maybe `ff_msgs` first?

marinagmoreira:
Hi all!
I think using jinja will make everything much more complicated and we should use minimalist strategies for the time being. For now I thin we should just rename the ros1 xml file and move on...

Here are somethings I've been doing this week (I should get a PR in today):
- Setting up ROS2 env: making the ros2 image from source smaller (existing images are 14+Gb which is unreasonable); layering gazebo + others on top (it doesn't come as standard on a from source build; setting up a functional devcontainer setup; setting up CI such that we don't have to build this natively (it's 350+ packages so docker takes forever...)
- Looking into simulation migration

Maybe we should start regular meetings to communicate more effectively and split tasks

marinagmoreira:
I found this thread that mentions agnostic .xml files https://github.com/ros-infrastructure/rosdep/issues/653
and I tested it using the same xml in ros1 and ros2 and works flawlessly, all we have to do is bump the package to format 3.

PR #565, opened 2022-11-03
Convert Simulink Controller to C++

bcoltin:
The basics are currently working in simulation. Further testing is needed in simulation and on the robot, then the old controller needs to be removed (currently it runs both and gives an error if they have a different output).


trey0:
It seems like a good idea to make the mode configurable (at run-time, without restarting FSW) between a few options like "old only", "use old but also compare with new and log errors as ROS telemetry", "use new but also compare with old and log errors as ROS telemetry", "new only". And make the default "old only" to start with.

Then this could be merged with `develop` and tested thoroughly, maybe even during facility CM sessions on the ISS, before switching to the new version for all users.

astrobee/config/gnc.config
```diff
@@ -261,9 +254,18 @@ hp_Ts = 1/62.5;
...
+tun_truth_q_omega_filt_cutoff_freq = 15*pi/180;
+tun_truth_q_omega_filt_enable = 0.0; -- Disabled by default
+tun_truth_velocity_filt_cutoff_freq = 15*pi/180;
+tun_truth_velocity_filt_enable = 0.0; -- Disabled by default
+
 -------------------------- SIM parameters ----------------------------
 
 tun_debug_ctl_use_truth = false; -- Controller will use ground truth instead of KFL estimate, only works in sim
+ctl_use_old = false;
+ctl_compare_old = false;
```
trey0, line 268: Great that these are in here! But I would have set `ctl_use_old` to `true`.

Rationale: It would make sense to cut a release to enable on-orbit testing of the new controller as an experiment during a facility CM activity. If that release had `ctl_use_old` enabled and `ctl_compare_old` disabled, its default behavior would be the same as the baseline and it could be used as the standard release. But we could also drop in a modified `gnc.config` for the experiment. Then we could update the flag defaults whenever the facility is comfortable with the new controller.

Another suggestion, if not already implemented, is to make sure these flag values are somehow recorded along with the rest of the telemetry. Just good for post-activity analysis sanity check, since it could otherwise be hard to figure out which controller was active.

bcoltin: I would suggest we start with the new one and revert if there is a problem. But regardless, the facility can decide and then set the value. For our own testing we should use the new one by default to see if there are problems so I would prefer to keep it as is.

bcoltin:
Tested in granite lab, succeeded at docking, undocking and square.

astrobee/config/gnc.config
trey0: I think we should change `ctl_use_old` to `true` prior to merging into `develop`.

If you want to set it to `false`, we should have a discussion with the facility first about the plan for testing and full-up deployment. This version shouldn't make it into an ISS release without that discussion.

bcoltin: Is there a reason our existing release testing procedures are insufficient to test this particular change? Do we have specific concerns that wouldn't be caught by them? Of course we should make the facility aware, but I don't see why this change means we should deviate from our normal procedure.

trey0: Some reasons a higher bar might make sense in this case: we're modifying code running on the LLP (supposed to be more stable) that has been stable and working well for a long time, breaking it could lead to unstable motion control which is a pretty bad outcome, and it's hard to justify the importance and urgency of the change from a stakeholder perspective, since there is no desired behavior change ("If it ain't broke, don't fix it.").

Maybe you have more evidence from prior testing than I'm aware of. Something that would help convince me is if we have logged ctl input and output from prior ISS activities and the validation checks pass on that data. But honestly, I would still suggest flying at least once on ISS with the two controllers running side-by-side with the old ctl in control and see that the checks pass, before letting the new ctl take over. It just seems like common sense.

P.S. If we are migrating other SIMULINK modules, is there ISS data that we don't have that would be valuable for testing? Maybe we can collect it on a future CM activity.

bcoltin: I'm open to different testing strategies when deploying on the ISS if we think it's necessary. We can discuss more on Monday.

But I strongly disagree about setting the default develop branch to use the old controller--- this will prevent us from testing the new controller internally, making a failure on the ISS more likely regardless of how we deploy.

marinagmoreira: If we're still worried even after bag testing, my suggestion would be to add either a service or a timer that reads the configs to be able to switch if during runtime

trey0: > I'm open to different testing strategies when deploying on the ISS if we think it's necessary. We can discuss more on Monday.

Sure.

> But I strongly disagree about setting the default develop branch to use the old controller--- this will prevent us from testing the new controller internally, making a failure on the ISS more likely regardless of how we deploy.

I don't see why it would prevent testing. It would just require a step in the testing procedure to temporarily modify the value of the flag in the config file before starting FSW.

I like the idea of being able to update the flag during execution, but recognize it might not be practical to implement. (Ctl is stateful, so you probably want to initialize both versions at the same time and keep them in lockstep. Perhaps an update when ctl is inactive due to idle propulsion would be ok. At any rate, it's not trivial.)

bcoltin:
Now tested and working on an ISS bag (did find one bug with using the second queued command). Once you review @marinagmoreira we can merge to develop based on our earlier discussion.

trey0:
> Now tested and working on an ISS bag (did find one bug with using the second queued command).

Not to slow things down, but I'm confused about the purpose of the last commit. It seems like it should have exactly the same behavior as the previous version, just using a pointer instead of a reference. Usually the difference between using a pointer and using an "equivalent" reference involves the possibility of NULL values, but that doesn't seem to apply here. I must be missing something.

bcoltin:
> > Now tested and working on an ISS bag (did find one bug with using the second queued command).
> 
> Not to slow things down, but I'm confused about the purpose of the last commit. It seems like it should have exactly the same behavior as the previous version, just using a pointer instead of a reference. Usually the difference between using a pointer and using an "equivalent" reference involves the possibility of NULL values, but that doesn't seem to apply here. I must be missing something.

That's what I thought too and how you would expect! This being C++, that is not the case--- with the first version, when we do curr = command_.current, we overwrite the value pointed to by curr, not change what it points to. Was a nasty bug to figure out.

https://isocpp.org/wiki/faq/references#assigning-refs

marinagmoreira (approved):

PR #569, opened 2022-11-08
Ros2 CI

marinagmoreira:


scripts/docker/ros2/ros2_rolling_base.Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+ENV ROSDISTRO=rolling
+ENV ROS_PYTHON_VERSION=3
+ENV ROS_VERSION=2
+
+SHELL ["/bin/bash", "-c"]
+ENV DEBIAN_FRONTEND=noninteractive
+
+# setup timezone
+RUN echo 'Etc/UTC' > /etc/timezone \
+    ln -s /usr/share/zoneinfo/Etc/UTC /etc/localtime \
```
trey0, line 12: Is there a missing `&&` before `ln -s`?

scripts/docker/ros2/ros2_rolling_deb.Dockerfile
```diff
@@ -0,0 +1,13 @@
+
+ARG UBUNTU_VERSION=16.04
```
trey0, line 2: I guess you are used to the `UBUNTU_VERSION` default value not mattering because it can be overridden from the wrapper script, but in this case I think choosing the value `16.04` is confusing because it is not one of the valid versions to use, correct? (Same comment applies to some other files.)

scripts/docker/build.sh
```diff
@@ -258,10 +270,26 @@ if [ "$astrobee_quick" = "true" ]; then
     build astrobee_quick "${revision}-" "quick-"
 fi
 
+if [ "$rolling_base" = "true" ]; then
+    build ros2/ros2_rolling_base "${revision}-" "rolling-base-"
```
trey0, line 274: In order to be kind to anyone who hopes to parse these image names later, maybe it would be better to not allow an internal hyphen in the last arg to `build`. Like call it `rollingbase-` or `rolling_base-` instead of `rolling-base-`? Not critical, just a suggestion.

scripts/docker/build.sh
```diff
@@ -130,10 +134,18 @@ while [ "$1" != "" ]; do
                                    ;;
         astrobee_quick )           astrobee_quick="true"
                                    ;;
+        rolling_base )             rolling_base="true"
```
trey0, line 137: The naming is a bit confusing here. One might think that `rolling_base` is a ROS2 version of the base Dockerfile such that you can build `astrobee` on top of it. But instead it's a base that you can build ROS2 on top of, like with `ros2_rolling_source.Dockerfile`, and then build `astrobee` on top of that.

I'm not sure what to suggest here exactly, but just noticing that as the number of targets in `build.sh` keeps expanding, it will get harder to understand how they're all supposed to relate to each other, and more apt names might help a bit.

trey0 (approved):
I had some detail comments.

Not sure what the intended status is after this commit? What is supposed to work now?

I appreciate the chance to review these changes, but we may not need to be so strict about our approval policy for this branch as long as it remains experimental.

scripts/docker/ros2/ros2_rolling_base.Dockerfile
marinagmoreira: nope, that actually copied from the official ros image and they do it like that, I thought it was weird too

scripts/docker/build.sh
marinagmoreira: Given that this is just the ros image from source, I think we'll have to add 2 more which is astrobee_ros2_base and astrobee_ros2 (note that astrobee_base builds on top of the noetic docker image). Once we're set on only one rolling/ros2 install method, this can compacted into one Dockerfile, but for now having the alternative is nice because it gives me some flexibility.

PR #570, opened 2022-11-08
Update code auto-generation scripts and docs workflow

trey0:
- Scripts support Python 3
- Enables running docs workflow under Ubuntu 20.04
- Enables faster `doxygen` install with `apt-get` (gets v1.8.17 instead of previous v1.8.20 source build, need to double-check output isn't wonky)

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -28,9 +28,9 @@ jobs:
     - name: Replace old documentation with new
       run: |
         (cd docs && git rm -r "*")
-        cp doc/documentation.html docs/index.html
-        cp doc/documentation.html docs/documentation.html
-        cp doc/README.md docs/README.md
+        cp repo/doc/documentation.html docs/index.html
+        cp repo/doc/documentation.html docs/documentation.html
+        cp repo/doc/README.md docs/README.md
         docker cp $(docker create --rm astrobee:documentation):/repo/doc/html docs/html
 
     - name: Commit and Push
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -27,13 +27,16 @@ jobs:
 
     - name: Replace old documentation with new
       run: |
-        [ -d docs ] || mkdir docs
-        [ -d docs/html ] && rm -r docs/html
+        (cd docs && git rm -r "*")
+        cp doc/documentation.html docs/index.html
+        cp doc/documentation.html docs/documentation.html
+        cp doc/README.md docs/README.md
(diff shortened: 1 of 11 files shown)


marinagmoreira:
The reason we froze the doxygen version and installed from source was because the output was wonky with apt-get :)

trey0:
> The reason we froze the doxygen version and installed from source was because the output was wonky with apt-get :)

Is it worth checking again under 20.04 if the `apt-get` version is more recent than 18.04? Or you already know it doesn't work right?

trey0:
> The reason we froze the doxygen version and installed from source was because the output was wonky with apt-get :)
> 
> (I remember the tree menu didn't parse correctly as well as some markup)

Now that the `gh-pages` setup on my fork is working,  I think this looks ok, but you can also check: https://trey0.github.io/astrobee/

One issue I had was that my `gh-pages` branch was previously in a weird state and the old CI workflow wasn't fully resetting the state and populating the root-level files on each docs rebuild. So I updated the workflow to do that, which should make things more repeatable for everyone.

I also had to make one tweak to the custom style sheet because the nav bar markup was very slightly different between doxygen versions. But I think now everything looks identical.

.github/workflows/docs.yaml
```diff
@@ -26,13 +27,16 @@ jobs:
 
     - name: Replace old documentation with new
       run: |
-        [ -d docs ] || mkdir docs
-        [ -d docs/html ] && rm -r docs/html
+        (cd docs && git rm -r "*")
+        cp repo/doc/documentation.html docs/index.html
```
marinagmoreira, line 31: I'm curious about why add this index.html that didn't exist previously

marinagmoreira (approved):
Looks good to me, I checked the html and looks identical

.github/workflows/docs.yaml
trey0: You can see the impact by comparing https://nasa.github.io/astrobee/ to https://trey0.github.io/astrobee/ . The former comes from GitHub not finding an `index.html` file and falling back to rendering the `README.md` file. That's not awful, but it seemed to me that any users who might land on the short URL would probably prefer to get redirected straight to the docs.

And actually, how I came to add it was a suggestion in the documentation for the GitHub Pages system that it uses the existence of an `index.html` file to infer what folders within the Pages branch should be offered as choices to become the documentation root. I'm not sure this is key, but it might help future developers who are activating the Pages feature for their fork. Mine was previously activated but in a weird broken state that took a while to figure out.

PR #573, opened 2022-11-09
ros2_migration: Astrobee package

marinagmoreira:
Open this PR for progress feedback and CI testing

### DIFF
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -2,6 +2,7 @@ name: Build and Test CI for Pull Requests
 
 on:
   pull_request:
+    branches: [ 'master', 'develop', 'ros2' ]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -2,7 +2,6 @@ name: Build and Test CI for Pull Requests
 
 on:
   pull_request:
-    branches: [ 'master', 'develop' ]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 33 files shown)


astrobee/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5)
```
trey0, line 18: Can this be moved inside the ROS2 branch of the conditional to avoid levying the greater version requirement on ROS1? Thinking of 16.04. Or fine if it's valid for 16.04, I didn't check.

.github/workflows/ci_pr.yml
```diff
@@ -2,7 +2,6 @@ name: Build and Test CI for Pull Requests
 
 on:
   pull_request:
-    branches: [ 'master', 'develop' ]
```
trey0, line 5: This seems like a big change, could trigger a lot of extra CI actions that people don't expect. The more conservative way to do it would be to just add `ros2`, no?

The `workflow_dispatch` option is always there so you can trigger it manually for any branch.

astrobee/CMakeLists.txt
marinagmoreira: I checked that my 16.04 cmake version was greater than 3.5

.github/workflows/ci_pr.yml
marinagmoreira: That was my initial thought too, but then I asked myself "will we ever want to open a pull request that doesn't have the build and test CI?". I do agree that it might trigger something extra that we don't want, I will revert this.

marinagmoreira:
I suppressed the python linter for now, will work on getting a linter that doesn't parse the entire thing and fails if it's different from expected output (similar to what we have in C++)

PR #575, opened 2022-11-15
Added wait parameter to restart gs command

kbrowne15:


astrobee/commands/freeFlyerPlanSchema.json
```diff
@@ -856,7 +856,13 @@
...
           "id": "apkName",
           "valueType": "string",
-          "notes": "Which guest science APK to terminae"
+          "notes": "Which guest science APK to restart"
+        },
+        {
+          "type": "ParamSpec",
+          "id": "wait",
+          "valueType": "long",
+          "notes": "The time in seconds the guest science manager waits in between sending the stop and start commands to the guest science apk."
```
trey0, line 865: It might be useful to provide some guidance about what value to use. Something like "Different apks will need different wait times to allow complete shutdown before starting again (or else risk crashing or hanging on restart). For the sci_cam_image apk which has to release the SciCam hardware resources, we empirically found a conservative wait time of X seconds worked reliably."

kbrowne15: That is a good idea! I'll add it now.

bcoltin (approved):

PR #579, opened 2022-11-17
Multibridge cleanup

rgarciaruiz:
* Remove unnecessary library from CMake file
* Remove unnecessary code from astrobee_astrobee_bridge.cc

### DIFF
--- a/communications/dds_ros_bridge/CMakeLists.txt
+++ b/communications/dds_ros_bridge/CMakeLists.txt
@@ -65,7 +65,7 @@ find_package(Soracore REQUIRED)
 
 
 catkin_package(
-  LIBRARIES dds_ros_bridge astrobee_astrobee_bridge
+  LIBRARIES dds_ros_bridge
   CATKIN_DEPENDS nodelet config_reader ff_util ff_msgs ff_hw_msgs dds_msgs
 )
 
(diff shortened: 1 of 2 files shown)


marinagmoreira:
@rgarciaruiz, this PR is flagged as Draft, is that accurate or is it ready for review?

rgarciaruiz:
I was planning to test this using both ground units first, however, since it actually works in simulation, the changes are very minor, and it does not affect immediate upcoming activities I say we can probably do that in the next software release. 
I'll go ahead and mark it as ready.

kbrowne15 (approved):

marinagmoreira:
> I was planning to test this using both ground units first, however, since it actually works in simulation, the changes are very minor, and it does not affect immediate upcoming activities I say we can probably do that in the next software release. I'll go ahead and mark it as ready.

Do we have a test scenario for astrobee_astrobee_bridge? Not sure if it's in the current release testing, but we can add to it :)

rgarciaruiz:
yeah, main issue is that you need to have two robots running FSW (and wannabee has Ubuntu 20, so we need to cross-compile for two Ubuntu versions). But, yes, we can start working on adding this test to the procedure, maybe as optional for now.

marinagmoreira:
@rgarciaruiz next time, make sure to squash merge!!!

rgarciaruiz:
> @rgarciaruiz next time, make sure to squash merge!!!

Yes, always, that's the option I chose, did something different happen?

marinagmoreira:
> > @rgarciaruiz next time, make sure to squash merge!!!
> 
> Yes, always, that's the option I chose, did something different happen?

Sorry I was too quick to reply :( the pr says "merged commit" so I thought it wasn't squash commit, but apparently it always says that regardless. I checked the commit history and everything looks great! (and proceeded to remove my comment)

PR #580, opened 2022-11-20
ROS2: ff_msgs and ff_hw_msgs

Pedro-Roque:
As in the title, porting ff_msgs and ff_hw_msgs. Moreover, the few changes were added:
1. Colcon Ignore to all packages except astrobee and communications
2. Deleted deprecated nodes


marinagmoreira:
Can we hold on deleting dds_msgs and dds_ros_bridge for now? At least until we test that the new comms work on the robot, Katie is looking into it

communications/ff_msgs/CMakeLists.txt
```diff
@@ -65,3 +66,63 @@ include_directories(
...
+find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(actionlib_msgs REQUIRED)
+find_package(trajectory_msgs REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+
+# Generate messages in the 'msg' folder
```
marinagmoreira, line 88: Can we find a way to not repeat this chunk of code?

trey0:
It's good to see some forward progress!

I agree with @marinagmoreira that, in general, we should try to be very conservative about changes that aren't 100% required to get things to build and run properly under ROS2. ("If it ain't broke, don't fix it.")

For example, it seems risky to change field names from camelCase to snake_case, even if it gets us closer to the usual ROS style conventions, because that change will affect lots of code throughout the repo, and even likely cause hassles with migrating legacy telemetry bags recorded on ISS. (Using ROS2 doesn't require snake_case, does it?)

trey0:
> At least until we test that the new comms work on the robot, Katie is looking into it

I'll take a stab at what conditions would need to be met before `dds_ros_bridge` is removed:
- We are past end-of-life for Astrobee ROS1 support (the ROS2 migration is fully complete for all legacy guest science on-orbit activities).
- AND we have some other solution for conveying ROS2-dialect DDS messages to the Astrobee control station, which currently speaks the NASA RAPID dialect of DDS.

Because of DDS dialects issue, it's quite possible that we'll continue using something analogous to `dds_ros_bridge` after the ROS2 transition, but that its job will change from bridging ROS1-TCPROS -> RAPID-DDS to bridging ROS2-DDS -> RAPID-DDS. And even if we find a workaround for the issue of different DDS dialects, a simpler bridge may still be worthwhile to help manage bandwidth on the space-to-ground link, for example by throttling message frequency on some topics.

communications/ff_msgs/action/Plan.action
```diff
@@ -28,7 +28,7 @@ float32 desired_omega                         # Desired (max) omega
 float32 desired_alpha                         # Desired (max) alpha
 float32 desired_rate                          # Desired rate
 
-duration max_time                             # Max generation time
+builtin_interfaces/Duration max_time                             # Max generation time
```
trey0, line 31: I don't think this version will compile in ROS1, so we would need a way to specify different field types for different ROS versions.

This becomes an example of where we might want to use the symlinking script, naming this file `jinja-Plan.action` and writing it:
```jinja
{% if USE_ROS1 %}
duration max_time
{% else %}
builtin_interfaces/Duration max_time
{% endif %}
```

Or maybe there is some other way to address it. If the build system doesn't require us to use the canonical folder and filename for each action file, I guess it can be handled by including different `Plan.action` files (perhaps in `ros1` and `ros2` subfolders) in the conditional branches of the `CMakeLists.txt`.

bcoltin: We all would like to make the ros2 branch ros1 compatible, but I'm not entirely convinced it will be feasible at a reasonable level of effort. I think we left it that we needed to better understand ROS2 and convert a first package before we would decide. Let's discuss tomorrow.

Pedro-Roque:
> trey0

Unfortunately it does @trey0, otherwise I would have skipped it. It errors out during the stage.

Regarding dds_ros_bridge, sounds good, I can keep it and let it be ignored for nowaybe?

communications/ff_msgs/action/Plan.action
Pedro-Roque: I think to make it compatible we need to go the template way, I'm not opposing it but am afraid of the extra effort needed

trey0:
> >  (Using ROS2 doesn't require snake_case, does it?)
> 
> Unfortunately it does @trey0, otherwise I would have skipped it. It errors out during the build stage.

I see. Then that's one more place to think about the best way to capture different versions.
 
> Regarding dds_ros_bridge, sounds good, I can keep it and let it be ignored for now maybe?

Well, it's already tracked separately in the task list at a lower priority, so I think it's the team consensus that it can wait.

communications/ff_msgs/action/Plan.action
trey0: A symlinking script that also implements templates is available at [1], tested and with usage instructions at the top. Probably it can be better tuned to agree with people's desired conventions and what we've learned since it was written. I don't insist that we use it and would be thrilled if we can find simpler approaches that make it unnecessary, but I am surprised to hear that it's not feasible!

[1] https://github.com/trey0/astrobee/blob/c4cafd764ddb9d80d4d5d52bb5cf6eacbb4f2d85/scripts/build/autogen_ros_version_src.py

Pedro-Roque: I'll give this a try and maybe bring some inputs by tomorrow's meeting @trey0

bcoltin: Feasible *at a reasonable level of effort*. I'm sure we can do it, but if it takes twice as much work as managing two branches, I don't think it's worth it.

Pedro-Roque:
> I see. Then that's one more place to think about the best way to capture different versions.
> 
> >

Trey, I'm not completely sure if I agree with going back to camelCase definitions if the ROS and Cpp best practice is lower case. I agree it is a breaking change, but maybe enforcing this would be a good way forward? I leave it to the team in any case https://google.github.io/styleguide/cppguide.html#Variable_Names

trey0:
> Trey, I'm not completely sure if I agree with going back to camelCase definitions if the ROS and Cpp best practice is lower case. I agree it is a breaking change, but maybe enforcing this would be a good way forward? I leave it to the team in any case https://google.github.io/styleguide/cppguide.html#Variable_Names

I'm not sure I understand exactly what you're proposing. We can discuss tomorrow.

My thinking was this: If we are able to figure out universal packages (ROS1+ROS2) I was assuming that the ROS1 version of the universal package would keep the legacy field names to avoid a breaking change in the guest science API, and the ROS2 version would grudgingly make whatever field name changes are required for ROS2 compatibility.

Code that uses the messages with modified field names could be well served by a definition like this:

```c++
#if ROS_VERSION == 1
#define FF_ROS_my_field_name myFieldName
#else
#define FF_ROS_my_field_name my_field_name
#endif
```
So then you could write `msg.FF_ROS_my_field_name` when you want to use the field, and it would work under both ROS versions. Annoying, but practical. And when ROS1 support goes end-of-life, we could more or less do a search-and-replace to get rid of the `FF_ROS_` prefix.

Pedro-Roque:
A probably good idea approach to ROS1/ROS2 compatibility:
https://github.com/dusty-nv/ros_deep_learning/blob/master/src/ros_compat.h

PR #581, opened 2022-11-21
Partition Image Files into Movement Sequences

rsoussan:
Changed the rotation detection script to partition a set of sequential image files around rotation/invalid movement.  This attempts to automate what we have been doing manually with each bagfile, splitting into sets of valid movements before running bundle-adjustment and merging.  The sequences are split around rotations and each sequence is saved in numeric order. Thus, valid sequences can be bundle-adjusted and merged first, and rotation sequences can optionally be added later when there is enough valid data in the map to handle them. Many of these changes stem from a discussion with Andres about how the tools are currently used for map creation.
The result can still be paired with removing low image movements.
@amoravargas


bcoltin (approved):
Good documentation :)

marinagmoreira (approved):

PR #582, opened 2022-11-22
ROS2: symlink building of ff_msgs

Pedro-Roque:
Here goes a build of ff_msgs with the symlink script. I ignored all packages except for ff_msgs. 

My take on it:
1. every time before `colcon build `we need to generate things - would be nice to automate this, not sure if we can specify project-based build rules that would automatically call the script whenever we do `colcon build`;
2. right now, files become `jinja-`, which is not good, because it's harder to look for them alphabetically. Maybe something like `abc.jinja.msg` is better;
3. a modification to a jinja file, even if not in the scope of jinja, still requires jinja to be updated (since the file will be replaced in-place under `src/`);
4. This one is easy to fix, but: `git_src` and `src` cannot be under the same workspace, otherwise `colcon` will complain of packages with repeated names, so `git_src` needs to a) be ignored by colcon, or b) moved to another folder (a hidden folder could probably work).

These were my findings. I'll see how it goes once I start porting the simulation package.

Tagging @marinagmoreira , @trey0 , @bcoltin 

I'll keep the other PR open so that we then decide which route to take.


trey0:
> /1. every time before `colcon build `we need to generate things - would be nice to automate this, not sure if we can specify project-based build rules that would automatically call the script whenever we do `colcon build`;
> /3. a modification to a jinja file, even if not in the scope of jinja, still requires jinja to be updated (since the file will be replaced in-place under `src/`);

I had the same thought but didn't yet investigate if there is an appropriate customizable pre-build hook to use so developers don't need to remember to run the script again.

> 2. right now, files become `jinja-`, which is not good, because it's harder to look for them alphabetically. Maybe something like `abc.jinja.msg` is better;

That sounds easy. Instead of `f.startswith("jinja-")` it could be `f.startswith("jinja-") or ".jinja" in f`. And likewise for `ros1-` and `ros2-`.

> 4. This one is easy to fix, but: `git_src` and `src` cannot be under the same workspace, otherwise `colcon` will complain of packages with repeated names, so `git_src` needs to a) be ignored by colcon, or b) moved to another folder (a hidden folder could probably work).

This might be a place where Astrobee's unusual folder conventions are causing confusion?

As I understand it:
1. A typical ROS workspace might have multiple source checkouts in a single workspace like:
```bash
~/ros_ws  # workspace
~/ros_ws/src  # where catkin/colcon does package discovery, checkouts in multiple subfolders often managed by vcs
~/ros_ws/src/std_msgs  # source checkout 1
~/ros_ws/src/trajectory_msgs  # source checkout 2
```
2. However, the Astrobee installation instructions recommend a simpler one-repo workspace structure which requires one less level of folders:
```bash
~/astrobee  # workspace
~/astrobee/src  # source checkout of astrobee repo, top level of checkout renamed to 'src' instead of 'astrobee'
```
I think if you're in case 2 the script works fine as written, at least under ROS1, because the `git_src` folder is not within the `src` folder used for package discovery. But if your workspace is structured like case 1, its path conventions don't make sense. It would be easy to add a feature to the script to support alternative path conventions. (And perhaps we should also reconsider our unusual recommended workspace structure.)

marinagmoreira (approved):
I'll open an issue with the raised concerns and merge this

Pedro-Roque:
> As I understand it:
> 
> 1. A typical ROS workspace might have multiple source checkouts in a single workspace like:
> 
> ```shell
> ~/ros_ws  # workspace
> ~/ros_ws/src  # where catkin/colcon does package discovery, checkouts in multiple subfolders often managed by vcs
> ~/ros_ws/src/std_msgs  # source checkout 1
> ~/ros_ws/src/trajectory_msgs  # source checkout 2
> ```

I think the issue is actually here @trey0 , since colcon for some reason doesn't do package discovery only on ` ~/ros_ws/src` but also on ` ~/ros_ws/git_src`. I might try to do a simple change of nameing to `src_git` and see what happens. Will report here again.

PR #584, opened 2022-11-23
ROS2: Symlinking for colcon build

Pedro-Roque:
Fixes some problems with having two srcs in the same folder.

To-Do/Fix:  figure out a way to run the symlink script automatically at build stage. I've found this:
https://colcon.readthedocs.io/en/released/user/configuration.html?highlight=colcon.pkg#colcon-pkg-files
...but didn't make something working with it. @trey0 in case you want to check?


scripts/build/autogen_ros_version_src.py
```diff
@@ -51,11 +52,19 @@
     boolean variable USE_ROS1 that can be used in Jinja2 conditionals to
     modify the content between ROS versions.
 This transformation has the following nice properties:
+<<<<<<< HEAD
```
marinagmoreira, line 55: merging shenanigans here :)

Pedro-Roque:
Fixed @marinagmoreira , ready for merge

marinagmoreira (approved):

PR #585, opened 2022-11-23
Ros2 migration description

marinagmoreira:


marinagmoreira:
tested that state_publisher runs when starting sim.launch.py

PR #588, opened 2022-11-25
exit or return in case pushd or popd fails

ajee10x:
Line 21:
"$DIR" Double quote to prevent globbing and word splitting.
Use 'pushd ... || exit' or 'pushd ... || return' in case pushd fails. 

Line 23:
Use 'popd ... || exit' or 'popd ... || return' in case popd fails.


bcoltin:
Thanks for your pull request. This code is actually slated for deletion in the next few months so we aren't going to change it at this point.

PR #589, opened 2022-11-28
ROS2 autogen script update

trey0:
Updates reflecting Pedro's comments and some other issues. (I did most of this before Pedro submitted #584  but was traveling Wednesday-Sunday and didn't get a chance to submit a PR until now.) Summary of changes:

- Treat a file as a Jinja2 template if it either starts with `jinja-` or contains `.jinja`. Same with `ros1` and `ros2`. Pedro's #584 did some of this but was less thorough.
- Change our Jinja2 rendering environment to strip whitespace around blocks by default so we don't need to worry about the `-` part of blocks like `{%- endif -%}`. This seems to be the preferred way to suppress unwanted whitespace without hassles.
- Switched the Jinja2 `USE_ROS1` flag to `ROS1`. It's shorter and agrees with the convention in the [ros_compat.h](https://github.com/dusty-nv/ros_deep_learning/blob/master/src/ros_compat.h) example.
- Which ROS version to use in the script is now controlled by the `$ROS_VERSION` environment variable or you can override with the `--ros-version` command-line flag. This is to agree with the most common convention for other tools.
- Made the checkout and workspace folders (previously hard-coded as `git_src` and `src`) configurable using command-line flags.
- Reworked Pedro's message changes to reflect the script updates above. Changed the `jinja-<name>.msg` files to `<name>.jinja.msg`. Changed `USE_ROS1` to `ROS1`. Got rid of the `-` in Jinja2 blocks.

This version keeps the same folder name convention as my original version but can be configured to use whatever folder names the developer likes. I wouldn't recommend the convention of using a dotfile `.astrobee` name as in Pedro's version because it will not appear in directory listings and seems likely to cause confusion, but would be open in general to changing the convention.


scripts/git/pre-commit.linter_python
```diff
@@ -43,16 +43,23 @@ if [ "$ans" -ne "0" ]; then
...
-    echo "Errors detected with 'black'. Fixing them. Try to add and commit your files again."
-    black . --include ${files}
-    failed_lint=true
+if false; then
+    echo "=================================================="
+    echo "  Analysing python code style with 'black'."
+
+    # Remove 'build' from black's default exclusions. We have
+    # a Python script in a 'build' folder.
+    default_black_exclude="/(\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|_build|buck-out|build|dist)/"
```
marinagmoreira, line 52: is this repeated on accident? (I don't see it being used

marinagmoreira (approved):

scripts/git/pre-commit.linter_python
trey0: No, it's just there to document the black default exclude string so you can see the only change we're making is to remove `build`. It could be commented out if that would be less confusing.

marinagmoreira (approved):

PR #591, opened 2022-11-29
ROS2: ff_common

Pedro-Roque:
This one also was rather easy to port. There was a nodelet dependency but this package didn't seem to contain any managed nodelets and seemed to be just a library.

PR #592, opened 2022-11-29
ROS2: conversion of config_reader

Pedro-Roque:
And config_reader, which for now complains about missing "ff_common/init.h" due to broken symlink in https://github.com/nasa/astrobee/pull/591 .


Pedro-Roque:
@marinagmoreira this branch should allow you to test it and get the same error that I get with the missing ff_common header

Pedro-Roque:
@trey0 Me and Marina were testing this out yesterday and compiling without the symlink worked fine but compiling with the symlinking didn't. Do you have the ROS2 build chain? It would be nice if you could test it as well. I believe the problem is on the path of the symlink done by the install() method in the CMakeLists, which possibly collides with the symlink generated with the script.

Pedro-Roque:
@marinagmoreira this also builds correctly, but check the change to CMakeLists.txt (I'm not sure where the version.txt is used, but maybe we can work on making it compatible with symlink builds.

trey0:
> @trey0 Me and Marina were testing this out yesterday and compiling without the symlink worked fine but compiling with the symlinking didn't. Do you have the ROS2 build chain?

Sad to say I didn't yet work on figuring out the ROS2 build chain. Partly I was hoping at some point there would be a step-by-step developer setup guide for Astrobee under ROS2. Does that exist and I missed it?

Based on the symptoms you described, I think the solution is as simple as removing the call to `os.path.relpath()` in the script, which causes it to write relative symlinks. I habitually prefer relative symlinks because they make it possible to relocate the whole workspace without breaking the links, but I don't think that's a major concern here. Absolute symlinks, on the other hand, can be copied to a new place without breaking... you just can't move the file the symlink points to.

Pedro-Roque:
Hmm I see your point, probably in this case it doesn't matter a lot, so using absolute paths would be just as good.

PR #596, opened 2022-11-29
pico_driver: document missing fields in point cloud metadata

### LINKED ISSUE #417: PointCloud2 messages from Pico Flexx should have "fields" metadata describing all fields of the point
Currently, the `fields` metadata in Pico Flexx PointCloud2 messages describes only the fields `x`, `y`, and `z`, all with float32 type. That adds up to only 12 bytes, but the `point_step` is 20 bytes.

Presumably the full 20-byte point is described by this data structure from libroyale (`/usr/include/royale/DepthData.hpp`):

    struct DepthPoint
        {
            float x;                 //!< X coordinate [meters]
            float y;                 //!< Y coordinate [meters]
            float z;                 //!< Z coordinate [meters]
            float noise;             //!< noise value [meters]
            uint16_t grayValue;      //!< 16-bit gray value
            uint8_t depthConfidence; //!< value from 0 (invalid) to 255 (full confidence)
        };

The field sizes in the struct add up to 19 bytes, and a final pad byte for 4-byte alignment would bring it up to 20.

Assuming this interpretation is correct, the missing data fields `noise`, `grayValue`, and `depthConfidence` should be added to the `fields` metadata, documenting their presence and making them easier to extract using standard point-cloud processing tools.

This is not a high-priority task, because nobody is actively using the fields. Ryan mentioned he might want to use the `depthConfidence` field to improve depth-based localization.

trey0:
Fixes #417.

Marked draft pending testing on the granite table. The test would be to record the relevant HazCam message topic, try to extract each newly documented field from the point cloud messages (ideally pack the values of the field into an image), and sanity check the results make sense (view the images).

@bcoltin what do you think? This doesn't seem urgent to test, just whenever we get around to it.

### DIFF
--- a/hardware/pico_driver/src/pico_driver.cc
+++ b/hardware/pico_driver/src/pico_driver.cc
@@ -231,23 +231,46 @@ class PicoDriverL1 : public PicoDriver, public royale::IDepthDataListener, publi
     cloud_.point_step = sizeof(struct royale::DepthPoint);
     cloud_.row_step = cloud_.width * cloud_.point_step;
     cloud_.data.resize(cloud_.row_step * cloud_.height);
-    // X, Y and Z
+
+    // Fill metadata about available fields in point cloud
     sensor_msgs::PointField field;
+
     field.name = "x";
     field.offset = offsetof(struct royale::DepthPoint, x);
     field.datatype = sensor_msgs::PointField::FLOAT32;
     field.count = 1;  // Number of ELEMENTS, not bytes!
     cloud_.fields.push_back(field);
+
     field.name = "y";
     field.offset = offsetof(struct royale::DepthPoint, y);
     field.datatype = sensor_msgs::PointField::FLOAT32;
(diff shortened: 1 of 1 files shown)


tejusp (approved):

bcoltin (approved):
We will still test this before merging.

marinagmoreira:
I can see that the fields are in the message:
---

- 
  name: "x"
  offset: 0
  datatype: 7
  count: 1
- 
  name: "y"
  offset: 4
  datatype: 7
  count: 1
- 
  name: "z"
  offset: 8
  datatype: 7
  count: 1
- 
  name: "noise"
  offset: 12
  datatype: 7
  count: 1
- 
  name: "grayValue"
  offset: 16
  datatype: 4
  count: 1
- 
  name: "depthConfidence"
  offset: 18
  datatype: 2
  count: 1
---


and I can import the recorded message to rviz with no issues and view the point cloud
so considering this tested

PR #597, opened 2022-11-30
autogen absolute symlinks; fix configure.sh missing step

trey0:
See discussion at #592


trey0:
Since #602 builds on this PR and seems to be getting close to merging, let's close this as a separate PR.

Issue #542, opened 2022-09-21
ROS2 Update


Pedro-Roque:
Just opening up this issue since I couldn't find anything regarding Astrobee and ROS2. Are there plans for the port? At KTH we are starting a new lab in which we plan to use Astrobee-analogues and we are planning to go for ROS2 from the start, so we would be interested in contributing to this effort (tagging @marinagmoreira @bcoltin @trey0 ).

trey0:
Hi, Pedro!

We are very interested in eventually migrating to ROS2, but there are a lot of complicated issues tied up with it.

A piece of good news is the Astrobee flight software team will have a small stream of new funding to study platform modernization, which can include considering both ROS2 and its Space ROS extensions. But I wouldn't expect to get to a fully supported ROS2 build for a while.

For one thing, our team is already burdened with supporting the software sim in Ubuntu 16.04, 18.04, and 20.04 (with their respective different ROS1 versions and third party library versions), so multiplying that sounds painful. There is an active effort preparing to migrate the Astrobee hardware units from 16.04 to 20.04, which should allow us to deprecate the older versions and simplify the code base a lot. It would be great if we could complete that simplification before investing seriously in ROS2 support.

There are also lots of open questions about how to do the migration. From my perspective, the theme of the migration should be to improve software quality by moving to the new framework whose core has a higher level of QA, and keeping the software stack as simple as possible for robustness and maintainability. I would probably advocate *for* moving to the most recent ROS2 version that runs on Ubuntu 22.04, due to its better maturity, even though that would make our overall ROS1-to-ROS2 jump bigger, and also *against* pursuing a hybrid build where ROS1 and ROS2 modules run side-by-side and communicate via a bridge, because I expect that might even be a step backward in terms of having more ways for things to go wrong. But these are all questions that need more study.

I think a realistic hope is to have a clear plan for this by say the end of February 2023. And it's great that you're raising the question now leading up to the next Astrobee Working Group meeting in October. We should probably use the AWG as an opportunity to ask about stakeholder needs and concerns relating to a possible migration. There's a lot of legacy code out there and we can't strand users who still have ISS activities left to run.

Something we hadn't considered is that outside developers might be able to provide a significant force multiplier for our team to help with the transition. That could make a big difference! So that's another place for more discussion. Thanks for raising this issue!

Pedro-Roque:
Hi @trey0!

[...]

PR #598, opened 2022-11-30
remove COLCON_IGNORE, instead rename package.xml to package.ros1.xml

trey0:
Builds on #597, should probably merge after that


trey0:
Relates to discussion in #542

trey0:
I realized there is a cleaner way to do this. We can add a feature to the autogen symlink script to use a `ROS2_IGNORE` file similar to `COLCON_IGNORE` but with our desired behavior for both ROS1 and ROS2. Closing this approach, will switch to that.

Why I like the new approach better: Nicer to have a smaller number of `ROS2_IGNORE` files at higher directory levels rather than forcing a rename of every `package.xml` file (almost 100 files).

Why name the marker `ROS2_IGNORE` rather than `COLCON_IGNORE`: the concept is similar but the implementation has subtly different behavior, so it's more transparent to give it a different name. That will also prevent any unexpected behavior in other places.

PR #600, opened 2022-12-01
ROS2: msg_conversions pkg + ff_util (discussion)

marinagmoreira:
I did this on top of the config_reader branch from @Pedro-Roque because that's a dependency

Things to discuss:
- Check logging migration template
`#include "rclcpp/rclcpp.hpp"`
`static const rclcpp::Logger LOGGER = rclcpp::get_logger("msg_conversions");`
`#define ROS_FATAL_STREAM(...)   RCLCPP_FATAL_STREAM(LOGGER, __VA_ARGS__)`
- Decide which migration template to use for messages:
    - Template used in msg_conversion.cc: `namespace sensor_msgs {typedef msg::PointCloud2 PointCloud2;}` or
    - Template used in the tools: `namespace GeometryMsgs = geometry_msgs;` (with subsequent bottom replacement of each instance)

Personally I prefer the first as it doesn't involve code change, only headers

ps: This branch compiles msg_conversions fully on my setup


shared/msg_conversions/tools/pose_stamped_msg_cnv.cc
```diff
@@ -48,12 +60,12 @@ void odometry_callback(geometry_msgs::PoseStampedPtr const & odometry) {
...
-  ros::init(argc, argv, "pose_stamped_msg_cnv");
+  ROS_CREATE_NODE("pose_stamped_msg_cnv");
+
 
-  ros::NodeHandle nh;
-  ros::Subscriber odometry_sub = nh.subscribe(FLAGS_input_topic, 5, &odometry_callback);
-  transform_pub.reset(new tf2_ros::TransformBroadcaster());
-  ros::spin();
+  auto odometry_sub = ROS_CREATE_SUBSCRIBER(geometry_msgs::PoseStamped, FLAGS_input_topic, 5, odometry_callback);
+  // transform_pub.reset(new tf2_ros::TransformBroadcaster());
```
marinagmoreira, line 67: not sure why this line is necessary and why it breaks the build on ROS2

shared/msg_conversions/tools/landmark_msg_cnv.cc
```diff
@@ -1,24 +1,59 @@
...
-ros::Publisher landmarks_pub;
+Publisher<sensor_msgs::PointCloud2> landmarks_pub = NULL;
+// auto landmarks_pub;
 
-void landmarks_callback(ff_msgs::VisualLandmarksPtr const & l) {
+void landmarks_callback(ff_msgs::VisualLandmarksPtr const l) {
   sensor_msgs::PointCloud2 out;
   out.header = std_msgs::Header();
-  out.header.stamp = ros::Time::now();
+  out.header.stamp = ROS_TIME_NOW();
```
trey0, line 56: This is another case using a macro that looks different from the idiomatic ROS2 API call, which I guess would be `node->getClock()->now()` (meaning you need to have the node pointer handy).

If the definition of `ROS_TIME_NOW()` is similar to the example in `ros_compat.h`, it's mildly concerning. Their `ROS_CREATE_NODE()` macro defines its own private global clock object to be used by future `ROS_TIME_NOW()` calls. This defeats the ROS2 concept, as I understand it, that each node has its own clock. I'm not clear how important that compartmentalization is, but if ROS2 supports running multiple nodes in the same process, it might be important to use the clock from the right node (e.g., some nodes might have their clocks initialized before others).

I think the namespaced inline function approach can't provide a definition that looks exactly like the ROS2 API call since the API involves a member function, but if we relax the requirement to match the syntax exactly, at least it could provide an inline function `ff_ros::now(node)` implemented as:
- ROS1: `ros::Time::now()` (ignoring the node argument)
- ROS2: `node->getClock()->now()`

Or we could decide to just stick with a macro for now, at the risk of having unexpected hassles later.

But if we are using macros for ROS abstraction, they should have a standard prefix such as `FF_ROS`...

shared/msg_conversions/src/msg_conversions.cc
```diff
@@ -18,90 +18,97 @@
 
 #include <msg_conversions/msg_conversions.h>
 
+#ifdef ROS1
 #include <ros/console.h>
+#else
+// Setup logging for ROS2
+#include "rclcpp/rclcpp.hpp"
+static const rclcpp::Logger LOGGER = rclcpp::get_logger("msg_conversions");
```
trey0, line 26: Following the pattern of making the code look like the ROS2 API (with a standard prefix), ideally we would want callers to write this:
```c++
auto logger = ff_ros::rclcpp::get_logger("msg_conversions");
if (crashed) {
  FF_ROS_RCLCPP_FATAL_STREAM(logger, ... stuff to log ...);
}
```
A sketch of how to support that:
```c++
namespace ff_ros {
#if ROS1
  namespace rclcpp {
    typedef Logger bool;
    inline Logger get_logger(const std::string& name) { return true; }
  }
#else
  namespace rclcpp = ::rclcpp;
#endif
}

#if ROS1
#define FF_ROS_RCLCPP_FATAL_STREAM(LOGGER, ...) ROS_FATAL_STREAM(__VA_ARGS__)
#else
#define FF_ROS_RCLCPP_FATAL_STREAM(LOGGER, ...) RCLCPP_FATAL_STREAM(LOGGER, __VA_ARGS__)
#endif
```

trey0:
> Decide which migration template to use for messages

I like the first case but suggest the resulting alias should be `ff_ros::sensor_msgs::msg::PointCloud2`. The abstraction might be defined as:
```c++
namespace ff_ros {
#if ROS1
  namespace sensor_msgs {
    namespace msg {
      typedef ::sensor_msgs::PointCloud2 PointCloud2;
    }
  }
#else
  namespace sensor_msgs = ::sensor_msgs;
#endif
}
```

Or this simpler version might be ok if it's the only thing we need from `ff_ros::sensor_msgs`:
```c++
namespace ff_ros {
#if ROS1
  namespace sensor_msgs {
    namespace msg = ::sensor_msgs;
  }
#else
  namespace sensor_msgs = ::sensor_msgs;
#endif
}
```

shared/msg_conversions/tools/pose_stamped_msg_cnv.cc
trey0: In ROS2 `TransformBroadcaster()` takes a required node argument.

This could be abstracted with a function `ff_ros::tf2_ros::get_transform_broadcaster(node)` that calls the `tf2_ros::TransformBroadcaster()` constructor with or without the node argument depending on ROS version. Or we could just write out the conditional each time if it's not called too often.

trey0: (Without the call `transform_pub` is basically a null pointer so any attempt to use it should crash.)

shared/msg_conversions/include/msg_conversions/msg_conversions.h
```diff
@@ -24,37 +24,51 @@
 #include <eigen3/Eigen/Core>
 #include <eigen3/Eigen/Geometry>
 
+#ifdef ROS1
```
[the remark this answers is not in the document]
marinagmoreira: doing this now, will clean up everything when final review

marinagmoreira:
Merged this PR and added some issues with pending problems.
Would also be great to revert the `add_definitions(-DROS1)` if we find a better alternative

PR #602, opened 2022-12-01
rename COLCON_IGNORE to ROS2_IGNORE, implement in autogen symlink script

trey0:
Builds on #597. That should be merged first.

This is the better approach that replaces #598.

Relates to discussion in #542.


scripts/configure.sh
```diff
@@ -344,7 +344,11 @@ if [ $skip_autogen == 0 ] ; then
     echo "running child instance of configure.sh in new autogen location..."
     "${autogen_path}/scripts/configure.sh" "${args_copy[@]}" -Z
 
-    echo "setting alias so ${build_cmd} runs autogen first on subsequent runs..."
+    echo "doing minimal build (just astrobee package) to force creation of devel/setup.sh..."
+    echo ${build_cmd} build astrobee
```
Pedro-Roque, line 348: This does not work on ROS2, here's the output:
```shell
➜  ros2_nasa_symlink_ws .astrobee/scripts/configure.sh -l -F -D

running autogen_ros_version_src.py to configure ROS version...
usage: autogen_ros_version_src.py [-h] [-v] [-r ROS_VERSION] [-d]
                                  [-c CHECKOUT_DIR] [-a AUTOGEN_DIR]
autogen_ros_version_src.py: error: autogen output dir /home/roque/Astrobee/src must not be a git checkout!
running child instance of configure.sh in new autogen location...
/home/roque/Astrobee/src/scripts/configure.sh: illegal option -- Z
configure.sh [-h] [-l <linux build>] [-a <arm build>] [-n <profile name>] [-p install_path] [-w workspace_path] [-B build_type] [-c <force clean cache>] [-C <don't clean cache>] [-d <with DDS>] [-D <without DDS>] [-r <with QP planner>] [-R <without QP planner>] [-f <with PicoFlexx driver>] [-F <without PicoFlexx driver>] [-k <use ccache if available>] [-K <do not use ccache>] [-v <with VIVE>] [-V <without VIVE>] [-t <with integration tests, requires gpu>] [-T <without integration test>] [-g <print debug information only>]
doing minimal build (just astrobee package) to force creation of devel/setup.sh...
colcon build astrobee
usage: colcon [-h] [--log-base LOG_BASE] [--log-level LOG_LEVEL]
              {build,extension-points,extensions,graph,info,list,metadata,test,test-result,version-check}
              ...
colcon: error: unrecognized arguments: astrobee
adding alias to devel/setup.sh so colcon runs autogen first on subsequent runs...
.astrobee/scripts/configure.sh: line 352: /home/roque/ros2_nasa_symlink_ws/devel/setup.sh: No such file or directory
(to suppress alias that invokes autogen, run \colcon instead of colcon.)
```

Pedro-Roque (changes requested):
The configure.sh script needs to be updated. I'll try to check for that.

The sourced directories also change: in ROS1, it goes to devel/, in ROS2 it goes to install/ .

scripts/build/autogen_ros_version_src.py
```diff
@@ -160,13 +169,19 @@ def get_desired_state(checkout_dir, autogen_dir, use_ros1):
...
     for dir_path, dir_names, file_names in os.walk(checkout_dir, followlinks=True):
-        if dir_path == checkout_dir and ".git" in dir_names:
-            dir_names.remove(".git")  # skip .git subfolder
+        if ignore_marker in file_names:
+            # if folder contains ignore_marker file, ignore it and all of its
+            # children
+            dir_names[:] = []
+            continue
+
+        if ".git" in dir_names:
```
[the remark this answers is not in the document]
trey0: Hm. I had a good reason to ignore the `.git` folder, which is that it's an error to try to use git inside the symlinked copy. I'm not sure what `git status` would show exactly, but there would be a bunch of bogus "changes" because the contents of the symlinked output are not the same as the source checkout by design, and it would break things if you accidentally commit some of these changes. So I thought it would be cleaner to leave out the `.git` folder so git would give you a helpful error that you are not inside a working copy.

There may be other files or folders where we don't benefit from symlinking them in the copy, but there's also no particular downside to having them there. In that case, I would say let's keep the script's ignore configuration simple (hard code only `.git` and the special ignore markers) even if it means the script does a bit of unnecessary extra work to link unneeded files.

But if you have an example in mind of other ignore patterns that would provide a significant benefit, that could change my opinion.

scripts/configure.sh
trey0: Great point. My assumption going into this PR was that `configure.sh` was still a ROS1-only script that needed to be migrated to support ROS2 in some way, and I wasn't trying to complete that ROS2 migration myself. Sorry if that was an invalid assumption.

Under ROS1, that line of the script runs `catkin build astrobee`, which is a valid command to build only the Astrobee package. I just guessed that the same command might work with `colcon`, but apparently not. Probably there is a simple way to get the same effect with `colcon`, in which case that exact step would work under ROS2, but then the rest of the ROS1-only `configure.sh` script would presumably break :) .

But I hope it's ok to merge this PR and complete the ROS2 port of `configure.sh` later.

Pedro-Roque:
Did a PR here: https://github.com/trey0/astrobee/pull/6  that fixes some of these, still I can't get `colcon` to run the autogen script on call.

trey0:
> Did a PR here: [trey0#6](https://github.com/trey0/astrobee/pull/6) that fixes some of these, still I can't get `colcon` to run the autogen script on call.

Based on what you said in trey0#6 I guess the various `setup.<SHELL>` scripts don't inherit from `setup.sh` anymore in ROS2, even though it worked in ROS1. Thanks for the multi-shell fix!

scripts/configure.sh
Pedro-Roque: I think this is fixed now?

PR #603, opened 2022-12-01
FAM Converted to C++

bcoltin:
Tested in simulation, bag from ISS and on granite table. Gives the same output as previous FAM. Doing merge and update the same way we did for the controller.


marinagmoreira (approved):

marinagmoreira:
I'm curious about whether we will continue using the gnc_autocode for things like this:
https://github.com/nasa/astrobee/blob/develop/simulation/src/gazebo_model_plugin_pmc/gazebo_model_plugin_pmc.cc#L467
or if it will also be migrated

bcoltin:
> I'm curious about whether we will continue using the gnc_autocode for things like this: https://github.com/nasa/astrobee/blob/develop/simulation/src/gazebo_model_plugin_pmc/gazebo_model_plugin_pmc.cc#L467 or if it will also be migrated

Planning to do that next and then once we test on ISS we can delete the autocode and matlab folders. The rest is simulation only so no need to test that on ISS.

gnc/fam/src/fam.cc
```diff
@@ -67,20 +100,185 @@ void Fam::CtlCallBack(const ff_msgs::FamCommand & c) {
...
+    uint8_t & impeller_speed_cmd, Eigen::Matrix<float, 6, 1> & servo_pwm_cmd,
+    Eigen::Matrix<float, 6, 1> & nozzle_theta_cmd,
+    Eigen::Matrix<float, 6, 1> & normalized_pressure_density, Eigen::Matrix<float, 6, 1> & command_area_per_nozzle) {
+  if (speed_gain_cmd > 3)
+    ROS_FATAL("Invalid speed command %d.", speed_gain_cmd);
+
+  Eigen::Matrix<float, 6, 1> discharge_coeff, nozzle_widths;
+  // Nozzle Discharge coefficient.  Determined via Test
+  if (is_pmc1)
+    discharge_coeff << 0.914971062, 0.755778254, 0.940762925, 0.792109779, 0.92401881 , 0.930319765;
```
trey0, line 205: Ok, this is disturbing. I don't know of any reason why the left and right prop modules would have systematically different nozzle discharge coefficients across different Astrobee hardware units, so probably these numbers represent measurements from the two prop modules of a specific Astrobee hardware unit long ago, and they were semi-accidentally propagated for use on all Astrobee units. I notice the fourth entry in particular varies quite a lot between the two modules.

I think you did the right thing here in terms of exactly matching the output of the SIMULINK implementation so we don't break anything by switching. But we should also probably file an issue to figure out how to set these parameters more accurately. Maybe we don't have real discharge coefficients for all the hardware units, but if I understand this correctly, it would probably make more sense to at least use the average coefficients from the old calibration for both prop modules rather than blindly assuming the left and right modules will differ in the same way across all units. Jesse might have more insight.

trey0 (approved):
One step closer to no SIMULINK!

I only had a couple of minor comments.

I'm also really curious to get your understanding of how the FAM currently responds to commands that saturate the propulsion, and whether you have any suggestions for improving that.

bcoltin:
> One step closer to no SIMULINK!
> 
> I only had a couple of minor comments.
> 
> I'm also really curious to get your understanding of how the FAM currently responds to commands that saturate the propulsion, and whether you have any suggestions for improving that.

I'd have to think about it some more. From what Nick told me from his prior testing, we can tell pretty easily that the commands have saturated the propulsion though...

gnc/fam/src/fam.cc
```diff
@@ -32,6 +34,37 @@ Fam::Fam(ros::NodeHandle* nh) : inertia_received_(false) {
   ReadParams();
   config_timer_ = nh->createTimer(ros::Duration(1), [this](ros::TimerEvent e) {
       config_.CheckFilesUpdated(std::bind(&Fam::ReadParams, this));}, false, true);
+
+  float units_in_2_m = 0.0254;
+  abp_PM1_P_nozzle_B_B << 6.00,  4.01, -1.56,  // [m]  PM1: Position vector of the nozzle locations in the body frame
```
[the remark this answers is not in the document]
bcoltin: Yeah it is for the next line, I'll change it.

And this is embarrassing but I actually didn't realize what in_2_m stood for when I converted it! I thought we were converting from who knows what into units of 2 meters...

gnc/fam/src/fam.cc
bcoltin: I was disturbed by this too. Agreed, let's stick to the old behavior for now.

PR #604, opened 2022-12-01
Better Nav Cam Timestamp

bcoltin:
Addresses #412 
Tested on granite table.


trey0 (approved):
Looks great! Thanks so much!

Issue #556, opened 2022-10-27
Docker updates + documentation


marinagmoreira:
Created this here to deal with docker upgrades

- In one place we recommend "docker exec -it ..." to run an arbitrary command in a container. In other places we recommend "./scripts/docker/run.sh --remote -- cmd ...". Are these different in some useful way (maybe we should explain how) or maybe we should consistently do it the same way in all examples?
- It's not explained that there is a 1-to-1 mapping between the available Dockerfiles and the build.sh targets. (So if you read the description of astrobee_quick and you're interested in using it, you would know to run build.sh astrobee_quick.)
- There isn't an example of how to use build.sh with the astrobee_quick target (like a "lifecycle" of first you build astrobee with this command, then you change something, then you build astrobee_quick with this command so now it's much faster)
- In the unit tests section, there is a discussion of how to use run.sh with --mount and do local build commands within the running container. Is this still a recommended approach or is it somehow redundant with astrobee_quick?

trey0:
- I think we agreed we should recommend to new developers working in software sim that they develop in Ubuntu 20.04 going forward. (The docs have the old recommendation to use 16.04. It's mentioned both in the installation overview and the non-NASA install instructions.)
- We should probably say explicitly that 22.04 is not supported, since at least one new user got confused about it.
- We should replace "the 64-bit version of Ubuntu" with "Ubuntu for amd64 architecture". More places are starting to support arm64 architecture and that could cause confusion.
- For complete novices, perhaps we should say: "There are two main profiles of the Ubuntu distribution, Desktop and Server. Either profile should work for Astrobee development. If you are new to Ubuntu, we recommend that you install the Ubuntu Desktop profile so you can use the full graphical user interface for your development."
- It might help to explain the terms "native build" and "cross-compile" (they appear as section headers, but the meaning might not be obvious). My brief version is: Our development environment is always amd64 architecture. Native build means building for amd64 so you can test in software sim on your development machine. Cross-compile means building for armhf architecture so you can run the code on the robot.

PR #608, opened 2022-12-09
Docker docs

trey0:
Relates to #556 

This started as writing some initial docs for VSCode inside a Docker container (probably could be improved a lot). In the process, I ended up doing some reorganizing so the flow made more sense. And provided more clarity about OS versions. And made lots of style fixes.

It's probably easiest to review if you look at the GitHub-rendered Markdown:
- https://github.com/trey0/astrobee/blob/docker_docs/INSTALL.md
- https://github.com/trey0/astrobee/blob/docker_docs/scripts/docker/readme.md

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,44 +1,48 @@
 # Install
 
-## Quick start using the remote Astrobee Docker images
+## System requirements
 
-*The following has been tested on native Ubuntu systems using X11 (the default). Please see [these ROS pages](http://wiki.ros.org/docker/Tutorials#Tooling_with_Docker) for more resources.*
+Ubuntu 20.04 is the preferred host OS for most Astrobee developers to use.
 
-Make sure you have Docker installed in your Ubuntu system following the [installation instructions](https://docs.docker.com/engine/install/ubuntu/) and [post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
+Here are the available host OS options with development roadmap details:
+- Ubuntu 20.04: This is the preferred host OS for most Astrobee developers to use. The Astrobee Facility team is currently preparing to upgrade the robots on ISS from Ubuntu 16.04 to Ubuntu 20.04, but we aren't yet ready to announce a deployment date for that upgrade.
+- Ubuntu 18.04: We are not aware of any current robot users that still need Ubuntu 18.04 support, and expect to discontinue support in the near future. New users should not select this host OS.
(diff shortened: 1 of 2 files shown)


scripts/docker/readme.md
```diff
@@ -1,104 +1,201 @@
...
+cd $ASTROBEE_WS/src
+./scripts/docker/run.sh --remote
+```
 
-## Run the Astrobee simulator in the container
+Invoking `run.sh --remote` calls Docker with the right arguments such that it will download the latest pre-built complete Astrobee Robot Software image from our official Docker image repository, start it as a container, and invoke a default command inside the container (which runs the software simulation).
+
+## Quick start build (if needed)
+
+The fastest way to recompile the ARS software to reflect your local changes is:
```
marinagmoreira, line 68: Why use `ARS`? We don't use this anywhere else and no explanation for the acronym is given. I suggest replacing it with `Astrobee` software (I see 15 instances to this in the PR, this comment applies to all)

INSTALL.md
```diff
@@ -1,44 +1,48 @@
...
-cd astrobee
-./scripts/docker/run.sh --remote
-```
+Your host OS must have an X11 server installed if you want to use graphical applications, even if you are developing inside a Docker container (the X11 application running inside the container will forward its interface to the host's X11 server). X11 comes with Ubuntu Desktop by default.
 
-For more options, see: \ref install-docker.
+If you plan to develop inside Docker, see [this page on using ROS with Docker](http://wiki.ros.org/docker/Tutorials#Tooling_with_Docker) for more details.
 
-## Building the code natively
+## Option 1: Install inside a Docker container
```
marinagmoreira, line 18: Technically not installing anything xD I suggest replacing with "Using", but optional

scripts/docker/readme.md
marinagmoreira: I just saw that apparently ARS is a thing xD Feel free to ignore this even if it makes it less readable imho

marinagmoreira (approved):

INSTALL.md
trey0: Oh, no, I disagree. This does download a pre-built Docker image to the local machine, which is a form of installation. Plus I felt like I should write it that way on the INSTALL page, after all :)

scripts/docker/readme.md
trey0: I appreciate this feedback. I'm torn because I don't like technical jargon, but "Astrobee Robot Software" is the actual name of our software package (title of the main README), and it also felt strange to keep writing it out all the time. Sometimes we informally write "FSW" which is a common aerospace acronym, but in my opinion too jargon-y if we fail to define it, and also not technically correct (lots of the stuff in the software package is not flight software at all).

So maybe it's a weird response, but my take on how to address your comment is to use proper technical writing style: introduce the acronym the first time we use it, and use it consistently (ironically, more often) after that.

PR #612, opened 2022-12-12
make ubuntu 18 and 20 supported with 20 preferred

marinagmoreira:


bcoltin (approved):
Thanks!

doc/general_documentation/INSTALL.md
```diff
@@ -2,18 +2,15 @@
 
 # Usage instructions for non-NASA users
 
-Install the 64-bit version of [Ubuntu 16.04](http://releases.ubuntu.com/16.04)
-on a host machine, and make sure that you can checkout and build code.
+Install the 64-bit version of [Ubuntu 16.04](http://releases.ubuntu.com/16.04),
```
trey0, line 5: This is an improvement on the old text, but I said something stronger in #608 (as a new user, don't start with 18.04 because we will end-of-life support for it soon).

In general, I think we should move anything that looks like "system requirements" to the main INSTALL page and have a pointer back to that from here ("remember to review the requirements" [link]). For example, the 4 GB of RAM I guess applies whether or not you're a NASA user or using Docker or whatever.

marinagmoreira: The 4GB barely applies to native FSW, if someone does that with Docker, is going to have a bad time xD

trey0: I think it would be great to have a bit more discussion about memory in the system requirements section.

If I understand correctly, 4 GB is the minimum for a bare-metal install and you would need more if you run inside a VM. Docker Desktop on Windows and Mac has to spawn a Linux VM to run its containers on, but plain Docker under Linux shouldn't have a big memory penalty. (It would be interesting if your experience conflicts with that.)

Since 4 GB is the minimum but more is better, it could be good also to mention the pain points. (I know the multi-threaded build often exhausts memory but there are workarounds. I don't run the simulator much so don't know if that tends to have problems, too.)

PR #622, opened 2023-01-04
set up astrobee_vnc docker example

trey0:
This PR is created as a working example for discussion.

Here's an overview: https://github.com/trey0/astrobee/blob/vnc/scripts/docker/readme_vnc.md

The point is to be able to run graphical applications on bartlett or other remote hosts and view the resulting display on your local host. This could be relevant for how we recommend future interns structure their dev environment!

This functionality may not belong as an extra `astrobee_vnc` layer on top of the normal `astrobee` images. It might be better to include it in the `astrobee` or `astrobee_base` Docker images instead. The current implementation optimizes for minimal changes and a quick build.

I didn't get to the point of trying all of the main applications we care about in the container. Marina's challenge was "show me xeyes", so here it is :)

### DIFF
--- a/scripts/docker/astrobee_vnc.Dockerfile
+++ b/scripts/docker/astrobee_vnc.Dockerfile
@@ -1,0 +1,34 @@
+# On top of the normal Astrobee image, this will set up a VNC server
+# running inside the Docker container so you can run graphical
+# applications and view them using a VNC client.
+
+ARG UBUNTU_VERSION=20.04
+ARG REMOTE=astrobee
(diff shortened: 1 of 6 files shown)


marinagmoreira:
This is really great!!! I can't test this atm (because I'm locked out of vpn with the expired badge), but I'm really exited for it :D

marinagmoreira:
closing this to merge https://github.com/nasa/astrobee/pull/744 that is based on this PR with added devcontainer functionality that allows the ssh and vnc server to be setup in one go instead of running multiple scripts

PR #627, opened 2023-01-09
Additional patches for Ubuntu 20

rgarciaruiz:
* Move Ubuntu 20 patches for external packages to a different directory
* Add patches for boringssl and fastboot
* Rebase cmake with the latest version for Ubuntu 20
* New cmake patches to allow compilation on chroot
* Add opencv patch that was omitted by mistake before
* Add script to build all debians on chroot without using APT, it works for amd64 too.


scripts/setup/install_desktop_packages.sh
```diff
@@ -50,12 +50,12 @@ then
...
-      sudo /bin/bash -c "echo \"deb-src http://astrobee.ndc.nasa.gov/software ${DIST} main\" >> $arssrc" || exit 1
+      sudo /bin/bash -c "echo \"deb [arch=amd64] http://astrobee.ndc.nasa.gov/software_new ${DIST} main\" > $arssrc" || exit 1
+      sudo /bin/bash -c "echo \"deb-src http://astrobee.ndc.nasa.gov/software_new ${DIST} main\" >> $arssrc" || exit 1
   else
       echo "Tunnelling to get the custom Debian"
-      sudo /bin/bash -c "echo \"deb [arch=amd64] http://127.0.0.1:8765/software ${DIST} main\" > $arssrc" || exit 1
-      sudo /bin/bash -c "echo \"deb-src http://127.0.0.1:8765/software ${DIST} main\" >> $arssrc" || exit 1
+      sudo /bin/bash -c "echo \"deb [arch=amd64] http://127.0.0.1:8765/software_new ${DIST} main\" > $arssrc" || exit 1
+      sudo /bin/bash -c "echo \"deb-src http://127.0.0.1:8765/software_new ${DIST} main\" >> $arssrc" || exit 1
       ssh -N -L 8765:astrobee.ndc.nasa.gov:80 ${username}m.ndc.nasa.gov &
```
trey0, line 59: For security reasons when you use the `-L` argument to `ssh` you should always start it with `127.0.0.1:`, which will restrict the local end of the SSH tunnel to bind to the loopback interface only.

Otherwise by default it will bind to all network interfaces on the local host, which you probably won't need, but it will make the tunnel accessible from other hosts on your local network. The tunnel will show up in a port scan and you might get a warning from network security admins.

rgarciaruiz: Interesting observation. However, since this is not exactly related to this PR and it was not introduced here maybe it can go on its own PR. What do you think?

Actually, do we still use the tunnel to get our debians? Is m still reachable without VPN?

trey0: Well my main point was to spread the word about the security issue. If you don't feel it makes sense to address it in this PR that's really up to you (I'm not trying to say it should be a blocker).

If it was me, I would probably add the `127.0.0.1:` prefix to immediately fix the security issue and also add a comment suggesting somebody check if the tunnel is still needed.

rgarciaruiz: Same here. We can keep the tunnel option for now just in case we cannot directly access the astrobee server for some reason. But maybe we should make NO_TUNNEL default?

debian/control
```diff
@@ -31,42 +31,43 @@ Description: Astrobee flight software configuration.
 Package: astrobee-comms
 Section: comm
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}
```
bcoltin, line 34: maybe should leave this in even if it's empty, in the hopes we may fix shlibdeps in the future?

debian/control
```diff
@@ -31,42 +31,43 @@ Description: Astrobee flight software configuration.
...
+Depends: ${misc:Depends}
   astrobee-config (>= ${binary:Version}),
   rti (>=1.0), libmiro0 (>=0.1), libsoracore1 (>=1.0),
-  ros-kinetic-ros-base, ros-kinetic-cpp-common,
-  ros-kinetic-common-msgs, ros-kinetic-nodelet, ros-kinetic-image-transport,
-  ros-kinetic-compressed-image-transport, ros-kinetic-tf2,
-  ros-kinetic-tf2-geometry-msgs, ros-kinetic-tf2-msgs, ros-kinetic-tf2-ros,
-  ros-kinetic-tf2-sensor-msgs,
+  ros-${ros-distro}-ros-base, ros-${ros-distro}-cpp-common,
```
bcoltin, line 37: nice!

scripts/build/build_debian.sh
```diff
@@ -1,19 +1,27 @@
...
 if [[ $* == *--config* ]]; then
   EXTRA_FLAGS="-A"
 fi
 
 pushd $DIR/../..
-export CMAKE_TOOLCHAIN_FILE=${DIR}/ubuntu_cross.cmake
-DEB_BUILD_OPTIONS="parallel=20" debuild -e ARMHF_CHROOT_DIR -e ARMHF_TOOLCHAIN -e CMAKE_TOOLCHAIN_FILE -e CMAKE_PREFIX_PATH -us -uc $EXTRA_FLAGS 
+DEBEMAIL="astrobee-fsw@nx.arc.nasa.gov" DEBFULLNAME="Astrobee Flight Software" dch -l"+$DIST" -D"$DIST" "Set distribution '$DIST' for local build"
+debuild -e ARMHF_CHROOT_DIR -e ARMHF_TOOLCHAIN -e CMAKE_TOOLCHAIN_FILE -e CMAKE_PREFIX_PATH -e ROS_DISTRO -e ROS_PYTHON_VERSION -us -uc $EXTRA_FLAGS
+git checkout debian/changelog
```
bcoltin, line 26: what is this for?

scripts/setup/debians/ubuntu20_patches/readme.md
```diff
@@ -0,0 +1,44 @@
+# Astrobee armhf debian patches for Ubuntu 20
+
+This folder contains scripts and patches needed for flight software on the
```
bcoltin, line 3: very nice, thank you!

bcoltin (changes requested):
Looks great! Main thing is I think we should move the packages to build the images to platform

debian/control
rgarciaruiz: I took it out because lintian complained about it. Maybe we can put it back once we make it work? Although, I'm not sure if it will ever work with xenial/focal and our chroot setup. dpkg has a new feature that can help us but it's not available for focal or xenial.

debian/rules
```diff
@@ -3,61 +3,59 @@
...
 endif
 
-# skip build for config files
-build-indep: ;
-	mkdir -p build
+# This is a Makefile variable that will only be used by native since armhf uses --extend
+CMAKE_PREFIX_PATH = $(CURDIR)/cmake:/opt/ros/$(ROS_DISTRO)
 
 %:
-	dh $@ --buildsystem=cmake --builddirectory=build/ --parallel
```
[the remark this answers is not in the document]
rgarciaruiz: Catkin should take care of that (it defaults to nproc). I believe that option is meant to be passed on to the build system, but catkin is not exactly supported anyway.

scripts/setup/debians/build_install_debians.sh
```diff
@@ -22,7 +22,7 @@
 set -e
 
 debian_loc=$(dirname "$(readlink -f "$0")")
-dist=$(grep -oP "(?<=VERSION_CODENAME=).*" /etc/os-release)
+dist=$(. /etc/os-release && echo $UBUNTU_CODENAME)
```
marinagmoreira, line 25: Do we still need this file with the build_debians.sh?

scripts/setup/debians/ubuntu20_patches/boringssl/changelog
```diff
@@ -0,0 +1,27 @@
+android-platform-external-boringssl (8.1.0+r23-2ubuntu1+focal1) focal; urgency=medium
```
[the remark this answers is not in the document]
rgarciaruiz: To be honest, I like them there since they fit nicely with the other Debian builds. I couldn't find a place where they would fit that well on astrobee_platorm. They are actually more related to our Debian repo.

marinagmoreira (changes requested):
I'll test this on my system asap and get back to you

scripts/setup/debians/build_install_debians.sh
rgarciaruiz: I think I didn't test it (build_debians.sh) in the context of users so I was planning to leave them both for a while and take care of that on a future PR. We would also need to update the docs. For users, I think it would be build_debians.sh --install-with-deps.
Do you think we should replace it already?

scripts/build/build_debian.sh
bcoltin: Just to be careful, might want to add a check before building that this file hasn't been modified since this would revert any changes

rgarciaruiz: The build_debian script had a check for that already, at the beginning. Do you think that's enough?

bcoltin: oh you're right! perfect then!

bcoltin (approved):
thanks, looks good once Marina does her testing!

scripts/setup/debians/ubuntu20_patches/boringssl/changelog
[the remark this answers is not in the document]
trey0: I think it's great to have as much of the software as possible available for external developers to see and don't see much downside to keeping it in `astrobee` vs. hidden in `astrobee_platform`.

However (I didn't check this), are we making it easy for people to build only the necessary packages? Whatever standard installation instructions we give to external users should probably not build the stuff @bcoltin is talking about.

marinagmoreira:
Had to:
 install: apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf
change toolchain file in platform condition to: if [ -e "toolchain/gcc/${MANIFEST}" ]; then

marinagmoreira (approved):
Good to merge after addressing:
Had to:
install: apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf (https://www.acmesystems.it/arm9_toolchain)
change toolchain file in platform condition to: if [ -e "toolchain/gcc/${MANIFEST}" ]; then

marinagmoreira:
The changes I referenced are both to be applied to submodules/platform, so I'll put it in a separate PR

Issue #561, opened 2022-11-01
Move DepthOdometry to mlp_depth_cam manager


trey0:
In [MLP.launch](https://github.com/nasa/astrobee/blob/develop/astrobee/launch/robot/MLP.launch#L92), the `DepthOdometry` nodelet is under the `mlp_vision` manager, where the NavCam images are published.

However, since its primary data source is the Pico Flexx depth camera point clouds and reflectance images, it may better belong under the `mlp_depth_cam` manager, where those messages are published.

This seems important (possibly a significant performance impact) but not urgent unless there is an ISS activity planning to use it.

PR #632, opened 2023-01-12
Fix depth odom nodelet manager

rsoussan:
Addresses issue 561: https://github.com/nasa/astrobee/issues/561


marinagmoreira:
This should be tested on the robot before merging because there are some configurations where it causes the fsw to crash at startup just by changing managers

trey0 (approved):
Looks good to me.

Marina has a good point about testing on the robot.

bcoltin (approved):
Yay, thanks!

tejusp (approved):

marinagmoreira:
Tested in the lab, started and stopped fsw 3 times, flew the robot around, all good

PR #633, opened 2023-01-12
Ros2 simulation setup

marinagmoreira:
Things done:
- Launch files related to simulation functional
- Robot + plugins being spawned + running
- Some gazebo plugins converted
- Other ROS2 fixes

Launch simulation with:
`ros2 launch astrobee sim.launch.py debug:=true rviz:=true`


shared/config_reader/CMakeLists.txt
```diff
@@ -19,8 +19,12 @@ project(config_reader)
 
 cmake_minimum_required(VERSION 3.5)
 
-## Compile as C++14, supported in ROS Kinetic and newer
-add_compile_options(-std=c++14)
+# Default to C++14
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
marinagmoreira, line 23: I changed it to this because as I was facing a crazy bug, one of the things that I tried was to use an Ubuntu 22 + humble docker image as base, and you need this for it to compile fine in that env (otherwise it seemed that it used C++17).

Also the good news is that I tested the entire pipeline until rn on humble and everything works exactly the same using this branch.

description/description/urdf/model_drag.urdf.xacro
```diff
@@ -20,7 +20,6 @@
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
   <gazebo>
     <plugin name="drag_ros" filename="libgazebo_model_plugin_drag.so">>
-      <namespace>/${ns}/</namespace>
```
marinagmoreira, line 23: soo one thing to look out for, if there is any variable inside the <plugin> tag, the plugin will crash when the command `gazebo_ros::Node::Get(sdf);` is executed. It took me a while to figure out the cause for the crash -.-

we'll have to figure out some other way to pass in the namespace to the plugin, I'll open an issue with this

mobility/framestore/tools/global_transforms.cc
```diff
@@ -69,7 +65,7 @@ int main(int argc, char **argv) {
           && msg_conversions::config_read_vector(&t_trans, &trans)) {
           // Only send global transforms
           if (global) {
-            tf.header.stamp = ROS_TIME_NOW();
+            tf.header.stamp = node_->get_clock()->now();
```
bcoltin, line 68: FF_TIME_NOW?

shared/config_reader/CMakeLists.txt
bcoltin: is there no way in catkin (or whatever it's called now) to define some flags over the entire build?

shared/config_reader/src/config_reader.cc
```diff
@@ -980,7 +980,7 @@ bool ConfigReader::ReadUInt(const char *exp, unsigned int *val) {
   bool ok = lua_isnumber(l_, -1);
   if (ok) {
     *val = static_cast<int>(rint(lua_tonumber(l_, -1)));
-    if (val < 0) {
+    if (rint(lua_tonumber(l_, -1)) < 0) {
```
bcoltin, line 983: ? why call this twice?

shared/ff_util/include/ff_util/ff_component.h
```diff
@@ -89,7 +89,7 @@ class FreeFlyerComponent {
 
   void AssertFault(FaultKeys enum_key,
                    std::string const& message,
-                   ros::Time time_fault_occurred = ROS_TIME_NOW());
+                   ros::Time time_fault_occurred = rclcpp::Clock().now());
```
bcoltin, line 92: FF_TIME_NOW?

simulation/package.xml
```diff
@@ -14,8 +14,12 @@
   <maintainer email="astrobee-fsw@nx.arc.nasa.gov">
     Astrobee Flight Software
   </maintainer>
-  <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
+  <build_depend condition="$ROS_VERSION == 1">catkin</build_depend>
```
bcoltin, line 17: didn't we decide to get rid of this, and use the ros2 in the ros2 branch and ros1 in develop?

bcoltin (commented):
Are we still planning to split things into ros2 and ros2-common, with the ros2-common shared between ros2 and ros1? It seems like this has things that should go into both ros2 and ros2-common.

shared/config_reader/src/config_reader.cc
marinagmoreira: because if you compare val in new compilers it complains that you're comparing an unsigned int with a signed int. This was a build error that poped up when compiling in ubuntu 22.
I can actually check if I can revert this after the change to the compile flags I did

shared/ff_util/include/ff_util/ff_component.h
marinagmoreira: Can't use this here because it needs to be static, which FF_TIME_NOW() isn't :/

mobility/framestore/tools/global_transforms.cc
marinagmoreira: So FF_TIME_NOW() uses the variable clock_ and not node_ :/ not sure what to do about these edge cases, it seemed weird to create clock_ just to call this given that it would be custom ROS2 code anyway, Maybe have a FF_TIME_NOW_NODE and FF_TIME_NOW that uses clock_? What do you think?

simulation/package.xml
marinagmoreira: this is actually something that could go to ros-common, unless you think it's too cluttered

shared/config_reader/src/config_reader.cc
trey0: The previous version of the code was actually broken. The check for a negative numeric value needs to come before the number is cast to an *unsigned* int. Casting to unsigned int potentially changes negative values to positive according to twos-complement representation rules, so the test becomes useless.

We should both move the negative value check before the cast and change the cast to `static_cast<unsigned int>`. Right now it is casting to a signed int then assigning that value to an unsigned int lval, which apparently is a valid implicit type conversion but is unnecessarily obfuscated.

As a side note, I think it's ill-advised in this context to accept any floating point value and round to the nearest integer. It would make more sense to complain about non-integer values. (Arguably this minimizes user surprise... if specifying a value of 3.6 doesn't raise an error they would probably not realize it just becomes 4 internally, which could cause a lot of confusion.) But I guess that would be an improvement to check carefully before merging.

astrobee/launch/sim.launch.py
```diff
@@ -132,6 +132,21 @@ def generate_launch_description():
         #   <!-- Connect and update environment variables if required -->
         #   <machine unless="$(eval arg('sim')=='local')" name="sim_server" default="true"
         #            address="$(arg sim)" user="astrobee" password="astrobee" timeout="10"/>
+        # Update the environment variables relating to absolute paths
+        SetEnvironmentVariable(name="ASTROBEE_CONFIG_DIR",
+                               value=os.getenv("ASTROBEE_CONFIG_DIR", "/home/astrobee/native/config"),
+                               condition=LaunchConfigurationNotEquals("sim", "local")),
```
kbrowne15, line 138: I think this is correct but I'm confused. So you are only setting the environment variable if sim doesn't equal local. Do we not need it set when the sim is local? Or are we requiring users to set it? Also, where did that default path come from? Does ROS2 not have a find package anymore?

kbrowne15 (changes requested):

description/description/urdf/model_laser.urdf.xacro
```diff
@@ -19,11 +19,11 @@
 
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
   <gazebo>
-    <plugin name="laser_ros" filename="libgazebo_model_plugin_laser.so">
+<!--     <plugin name="laser_ros" filename="libgazebo_model_plugin_laser.so">
```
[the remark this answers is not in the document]
marinagmoreira: https://github.com/nasa/astrobee/issues/635

simulation/src/gazebo_model_plugin_truth/gazebo_model_plugin_truth.cc
```diff
@@ -92,27 +98,27 @@ class GazeboModelPluginTruth : public FreeFlyerModelPlugin {
...
+    pub_truth_pose_ = nh->create_publisher<geometry_msgs::PoseStamped>(
       TOPIC_LOCALIZATION_TRUTH, 1);
-    pub_truth_twist_ = nh->advertise<geometry_msgs::TwistStamped>(
+    pub_truth_twist_ = nh->create_publisher<geometry_msgs::TwistStamped>(
       TOPIC_LOCALIZATION_TRUTH_TWIST, 1);
 
     // Called before each iteration of simulated world update
-    timer_ = nh->createTimer(ros::Rate(rate_),
-      &GazeboModelPluginTruth::TimerCallback, this, false, true);
+    timer_ .createTimer(1 / rate_,
```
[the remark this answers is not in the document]
marinagmoreira: good eye!

astrobee/launch/sim.launch.py
marinagmoreira: So that is me converting this from astrobee.launch, line 130 to ROS2:
`<env unless="$(eval arg('sim')=='local')"`
if the sim is local it will use the:
`SetEnvironmentVariable(name="ASTROBEE_CONFIG_DIR",    value=os.getenv("ASTROBEE_CONFIG_DIR",    get_path("config"))),`
that was defined in this file previously

The find package now changed to be the function `get_package_share_directory(package)`. This is now the only thing available to my knowledge and points to the install folder, not build or src folder, so there is the added nuance that everything needs to be installed when launching

marinagmoreira: Also I'm not sure where those paths come from, they pre-date me xD

kbrowne15: Yeah, I was just confused. Thanks for clarifying.

shared/config_reader/CMakeLists.txt
marinagmoreira: https://github.com/nasa/astrobee/issues/636

shared/ff_util/include/ff_util/ff_component.h
marinagmoreira: https://github.com/nasa/astrobee/issues/637

shared/config_reader/src/config_reader.cc
marinagmoreira: I moved the check to before the assignment, if there's further desired changes, let's do it in a separate PR

trey0: (I don't understand. It didn't look like you changed anything since the comments.)

mobility/framestore/tools/global_transforms.cc
bcoltin: It's probably not too important, but why make the clock_ variable at all? We could just call get_clock in the macro. I doubt get_clock is an expensive operation?

marinagmoreira: Yeah it was one of the things that I didn't have a preference for and chose randomly whether to store a node_ variable or clock_ variable :)

PR #641, opened 2023-01-20
Ros2 pmc gazebo model plugin

marinagmoreira:


communications/ff_msgs/msg/AccessControlStateStamped.msg
```diff
@@ -15,7 +15,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 #
-# State of the access control node. Loosely based off of AccessControlState.idl 
+# State of the access control node. Loosely based off of AccessControlState.idl
```
marinagmoreira, line 18: These were prompting lint error on humble

simulation/CMakeLists.txt
```diff
@@ -77,6 +78,7 @@ find_package(JsonCpp REQUIRED)
 include_directories(
   include
   ${config_reader_INCLUDE_DIRS}
+  ${pmc_INCLUDE_DIRS}
```
marinagmoreira, line 81: need to figure out why this was needed, I set the ament_export_include_directories and added it to the ament_target_dependencies, but it still complained that it couldn't find #include <pmc/pmc_sim.h>

gnc/pmc/include/pmc/fam.h
```diff
@@ -21,6 +21,8 @@
 
 #include <Eigen/Dense>
 
+#include <ff_common/ff_ros.h>
```
bcoltin, line 24: we shouldn't need this here. put in C++ file to reduce compile times

gnc/pmc/CMakeLists.txt
```diff
@@ -15,30 +15,26 @@
...
+cmake_minimum_required(VERSION 3.5)
 project(pmc)
 
-## Compile as C++14, supported in ROS Kinetic and newer
-add_compile_options(-std=c++14)
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+set( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wall -O3 -fPIC" )
```
bcoltin, line 25: didn't we used to have some extra flags, like WError?

gnc/pmc/CMakeLists.txt
```diff
@@ -15,30 +15,26 @@
...
 
-## Find catkin macros and libraries
-find_package(catkin2 REQUIRED COMPONENTS
-  roscpp
-  config_reader
-)
+# Build for ROS2
+add_definitions(-DROS2)
```
bcoltin, line 28: I thought we said we wouldn't do this, we would just define it once in the ff_ros.h header for our branch?

simulation/CMakeLists.txt
bcoltin: I'm reading in the ament documentation that to export the header files we should use this:

```
  target_include_directories(my_target
    PUBLIC
      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
      $<INSTALL_INTERFACE:include>)
```

https://docs.ros.org/en/humble/How-To-Guides/Ament-CMake-Documentation.html

Might want to try that in the pmc CMakeLists

bcoltin (changes requested):

gnc/pmc/CMakeLists.txt
bcoltin: ignore this for now, I will make a separate pull request

gnc/pmc/CMakeLists.txt
marinagmoreira: I copied this from a different package, I see that the examples have -Wall -Wextra -Wpedantic, I think I'll do a separate PR with all the packages updated. (The -fPIC I added because it's needed to export a library)

gnc/pmc/include/pmc/fam.h
marinagmoreira: now we need it because of the logger macro :). But the config_reader was a dependency already, which depends on ff_common, so it shouldn't be too bad

kbrowne15 (approved):

bcoltin (approved):
Move include like we talked about and then looks good!

PR #642, opened 2023-01-20
Update monitor plugins and fix heartbeat name bug.

bcoltin:
Please make sure I'm not breaking anything else with the changes to ff_component.


marinagmoreira:
> Please make sure I'm not breaking anything else with the changes to ff_component.

That function is only used for gazebo plugins, so you're good!

marinagmoreira (approved):
It's awesome that we reduced the amount of plugins!

marinagmoreira:
I also tested this in humble and all the heartbeats were being printed well with `ros2 topic echo /heartbeat`

kbrowne15 (approved):
Great idea on the heartbeat plugin!

PR #643, opened 2023-01-20
Log creation macro.

bcoltin:
I thought the log creation would be less confusing with a macro. It was confusing because we had to define a specific variable which we then never referenced. It was also a common name likely to cause conflicts. This way still has its own problems of course...

What do you think?


marinagmoreira (approved):

marinagmoreira:
> I thought the log creation would be less confusing with a macro. It was confusing because we had to define a specific variable which we then never referenced. It was also a common name likely to cause conflicts. This way still has its own problems of course...
> 
> What do you think?

I agree that if we're hiding the LOGGER variable usage in the FF_ERROR(etc), it's confusing to define it. This way it's explicit that it's used for the purpose of FF_(...).

PR #646, opened 2023-01-21
Ros2 sparse map gazebo plugin

marinagmoreira:
Still WIP:

- [x]  need to add the opencv install to the CI (I tested it by installing an opencv debian that I had)
- [x] convert the rest of the camera package


astrobee/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ install(DIRECTORY launch
 
 # Mark resource files for installation
 install(DIRECTORY resources
-  DESTINATION lib
+  DESTINATION share/${PROJECT_NAME}
```
marinagmoreira, line 60: need to change this such that we can find the resources folder from the launchfile

shared/ff_util/include/ff_util/ff_timer.h
```diff
@@ -48,26 +48,30 @@ class FreeFlyerTimer {
...
   ~FreeFlyerTimer() {}
 
-  void createTimer(double period_sec,
+  void createTimer(double period,
                    TimerCallbackType cb_timer,
                    rclcpp::Node::SharedPtr node,
                    bool one_shot = false,
                    bool auto_start = true) {
-    createTimer(std::make_shared<rclcpp::Duration>(period_sec, 0),
+    int period_sec = std::floor(period);
```
marinagmoreira, line 56: @kbrowne15 I was getting a really high frequency and I think it's because it was rounding the seconds to zero, this fixed it

localization/camera/CMakeLists.txt
```diff
@@ -54,38 +50,41 @@ catkin_package(
 # Specify additional locations of header files
 include_directories(
   include
-  ${catkin_INCLUDE_DIRS}
   ${EIGEN3_INCLUDE_DIR}
   ${OpenCV_INCLUDE_DIRS}
+  ${LUAJIT_INCLUDE_DIR}
```
bcoltin, line 55: do we know why these aren't brought in by the config_reader dependency?

localization/camera/CMakeLists.txt
```diff
@@ -54,38 +50,41 @@ catkin_package(
...
+ament_export_targets(camera HAS_LIBRARY_TARGET)
+target_link_libraries(camera ${EIGEN_LIBRARIES} ${OpenCV_LIBRARIES})
 
-## Declare a C++ executable: undistort_image
+# Declare a C++ executable: undistort_image
 add_executable(undistort_image tools/undistort_image.cc)
-add_dependencies(undistort_image ${catkin_EXPORTED_TARGETS})
 target_link_libraries(undistort_image
-  camera gflags glog ${catkin_LIBRARIES})
+  camera gflags glog -lboost_filesystem -lboost_system)
```
bcoltin, line 70: should use Boost_LIBRARIES

bcoltin (changes requested):

localization/camera/CMakeLists.txt
marinagmoreira: It complained about not finding it on a piece of code in the camera package that imported a lua header, so not through the config_reader. I'm not sure why it didn't get it through the config reader..

localization/camera/CMakeLists.txt
marinagmoreira: interesting, I tried that first and didn't work...

localization/camera/CMakeLists.txt
bcoltin: you also need to add filesystem to the boost components that's probably why

localization/camera/CMakeLists.txt
bcoltin: you also need to add filesystem to the boost components that's probably why

marinagmoreira: worked like a charm!

localization/camera/CMakeLists.txt
bcoltin: actually, sorry, we shouldn't need this at all. Could you move all the includes here to the .cpp file? https://github.com/nasa/astrobee/blob/ros2/shared/config_reader/include/config_reader/config_reader.h 

You can forward declare the lua_State struct

shared/ff_util/include/ff_util/ff_timer.h
kbrowne15: Okay, thanks for fixing it.

kbrowne15 (approved):

localization/camera/CMakeLists.txt
marinagmoreira: ups, I deleted those from the cmake and it worked (compiled from scratch and ran the tests), I think some other PR might have fixed the linking problem

bcoltin (approved):
Thanks!

Pedro-Roque:
@marinagmoreira this is not compiling on the VM - I'll build/install the opencv that's bundled with Astrobee and see if that fixes it.

PR #648, opened 2023-01-24
ROS2: Dock+Nav Cam plugin

Pedro-Roque:
First take at the navcam plugin.


marinagmoreira:
can you pull from the nasa/ros2 branch? We have those CI actions updated

Pedro-Roque:
Same here. I probably should have done a rebase so that you don't see previous commits there though

Pedro-Roque:
Fixed the history and everything compiles - also added the dock cam plugin to this

bcoltin:
So, the callbacks for changes in subscribers are dropped in ros2. This means the ToggleCallback in the current version is only called intiailly, and if there are no subscribers, I believe the images will never be published, even if subscribers are added later.

I guess we need to change this to always publish? Or stick a timer to call ToggleCallback repeatedly?

What do you all think?

marinagmoreira (approved):

marinagmoreira (approved):

bcoltin (approved):
Thanks!

Pedro-Roque:
Should this be merged? I'll work on top of it for the next camera ports

PR #658, opened 2023-01-30
ROS2: Optical Flow camera

Pedro-Roque:
Adding support for Optical Flow camera. 

Needs:
1. libopencv3.3.1 and libopencv-dev3.3.1 (used locally built ones, building in docker is not working atm);
2. libceres-dev (install manually in docker with apt install libceres-dev, we should bring this to the docker image).

Other than that, seems to be ready! Builds on top of https://github.com/nasa/astrobee/pull/648

PR #659, opened 2023-01-30
ROS2: AR Tags 

Pedro-Roque:
Adding support for AR Tags on ROS2. Everything went smoothly after https://github.com/nasa/astrobee/pull/658 , on which this PR relies.


bcoltin:
I think it will be much easier if you start each PR from the main branch instead of from the previous PR

Pedro-Roque:
> I think it will be much easier if you start each PR from the main branch instead of from the previous PR

I probably should have done that, I was hoping that as each PR gets in, I would be able to pull from ros2 and fix things hehe.

Pedro-Roque:
Closing this PR as the content will be shipped within https://github.com/nasa/astrobee/pull/660

PR #661, opened 2023-01-30
updating simulation documentation for ROS2

marinagmoreira:


scripts/docker/readme.md
```diff
@@ -48,6 +48,24 @@ You can manage your Dev Containers configuration using the files in the `.devcon
 
 You can start by selecting `View->Terminal` in the VSCode graphical interface. This will display a terminal session inside the Docker container where you can run arbitrary commands. Your container will persist throughout your VSCode session, and changes you make using the VSCode editor will be reflected inside the container, making it easy to do quick interactive edit/build/test cycles.
 
+## Enable x-forwarding from the Dev Container
+
+In a cmd line in your system run:
+```bash
+xhost local:docker
```
bcoltin, line 55: thank you :)

bcoltin (approved):

scripts/docker/readme.md
trey0: I think this particular new section is equally useful for the `develop` branch, right? In general, if we notice that in time, I would suggest making the change in `develop` first, then merging to `ros2`, as the easiest workflow.

(The technical writer inside me also wants to make some copy edits concerning capitalization and SMS-speak abbreviations, but I'll always have a chance to do another pass later if you don't feel like it right now.)

PR #663, opened 2023-01-31
ros2 ff action

kbrowne15:
This pr contains the ros2 ff action work and the nominal behavior unit test for the action. The rest of the unit tests will be finished in another pr.


bcoltin (approved):
Somehow I had never realized our actions were so complicated. Maybe you could convert a package that uses the actions next, after you finish the tests?

marinagmoreira (approved):

PR #660, opened 2023-01-30
ROS2: Optical Flow, AR Tags, Handrail Detect

Pedro-Roque:
Adding handrail detect plugin.

This plugin had not been ported to Gazebo > 7. Probably it's not being used anymore? I ported the content to the current version of Gazebo and everything compiles fine.

PR #666, opened 2023-02-02
ROS2 Ground Truth Localizer

marinagmoreira:
…ime a function


astrobee/launch/robot/MLP.launch.py
```diff
@@ -59,13 +59,12 @@ def generate_launch_description():
...
-            #     plugin='ground_truth_localizer::GroundTruthLocalizerNodelet',
-            #     name='ground_truth_localizer',
-            #     remappings=[('/image', '/burgerimage')],
-            #     parameters=[{'history': 'keep_last'}],
-            #     extra_arguments=[{'use_intra_process_comms': True}]),
+            ComposableNode(
+                package='ground_truth_localizer',
+                plugin='ground_truth_localizer::GroundTruthLocalizerNodelet',
+                name='ground_truth_localizer',
+                # condition=IfCondition(LaunchConfiguration("gtloc"))
```
marinagmoreira, line 66: So the `ComposableNode` doesn't support the `condition` arg, will open a PR trying to solve this separately -.-

marinagmoreira:
Will wait until https://github.com/nasa/astrobee/pull/660 is merged such that I can apply the clock_ fix to it

localization/localization_common/src/ros_timer.cc
```diff
@@ -21,15 +21,15 @@
 
 namespace localization_common {
 RosTimer::RosTimer(const std::string& timer_name) : averager_(timer_name, "time", "seconds"), start_time_(0) {}
-void RosTimer::Start() { start_time_ = ros::Time::now(); }
+void RosTimer::Start() { start_time_ = rclcpp::Clock().now(); }
```
marinagmoreira, line 24: Here because it is a time diff I used the static, but we can also add a NodeHandle field to the function to get the time if that makes a difference. What do you think?

shared/ff_util/include/ff_util/ff_component.h
```diff
@@ -98,11 +98,12 @@ class FreeFlyerComponent {
   std::string GetName();
   std::string GetPlatform();
   std::string GetTransform(std::string const& child);
+  rclcpp::Time GetTimeNow();
 
  protected:
   // Virtual methods that *can* be implemented by FF nodes. We don't make
   // these mandatory, as there is already a load callback in Gazebo.
-  virtual void Initialize(NodeHandle node) {}
+  virtual void Initialize(NodeHandle &nh) {}
```
marinagmoreira, line 106: Changed this such that the initialize function looks the same in plugins and components (old nodelets)

localization/ground_truth_localizer/package.xml
```diff
@@ -13,18 +14,28 @@
   <maintainer email="astrobee-fsw@nx.arc.nasa.gov">
     Astrobee Flight Software
   </maintainer>
-  <buildtool_depend>catkin</buildtool_depend>
-  <build_depend>roscpp</build_depend>
+  <build_depend condition="$ROS_VERSION == 1">catkin</build_depend>
```
bcoltin, line 17: I thought we were getting rid of this

localization/localization_common/CMakeLists.txt
```diff
@@ -15,39 +15,35 @@
...
-  INCLUDE_DIRS include 
-  LIBRARIES ${PROJECT_NAME} ${GTSAM_LIBRARIES} gtsam
-  CATKIN_DEPENDS
-    camera
-    config_reader
-    msg_conversions
-    ff_msgs
-)
+# Find OpenCV
+SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/../../cmake")
```
bcoltin, line 45: usually better practice to append rather than replace

localization/localization_common/src/ros_timer.cc
bcoltin: I think the way you have it now is better, doesn't seem worth the trouble

shared/msg_conversions/tools/landmark_msg_cnv.cc
```diff
@@ -48,12 +48,12 @@ DEFINE_string(output_topic, "/loc/rviz",
               "The output topic suitable for rviz.");
 
 Publisher<sensor_msgs::PointCloud2> landmarks_pub = NULL;
-std::shared_ptr<rclcpp::Clock> clock_;
+NodeHandle node_;
```
bcoltin, line 51: The _ is supposed to be for member variables only

shared/msg_conversions/tools/pose_stamped_msg_cnv.cc
```diff
@@ -45,11 +45,11 @@ DEFINE_string(output_frame, "ground_truth",
               "The frame to output to tf2.");
 
 std::shared_ptr<tf2_ros::TransformBroadcaster> tf_broadcaster_;
-std::shared_ptr<rclcpp::Clock> clock_;
+NodeHandle node_;
```
bcoltin, line 48: same

bcoltin (changes requested):
Looks good! Just some very minor suggestions

localization/localization_common/src/ros_timer.cc
trey0: It depends on the purpose of the timer. If it's only intended to measure wallclock time, seems great as-is. But if it's ever intended to measure sim time when running in sim (might not be at 1:1 speed), the other way might be needed.

bcoltin: oh, I hadn't realized this would be wallclock. yeah we probably want to get the ros time(?)

bcoltin (approved):
Oops, missed this somehow! Looks good thanks!

PR #671, opened 2023-02-04
replace list with append in CMakeLists

marinagmoreira:


bcoltin (approved):
Oops, should have noticed this earlier! Thanks!!

PR #674, opened 2023-02-07
Unlatching topics

marinagmoreira:


shared/ff_util/src/ff_nodelet/ff_nodelet.cc
```diff
@@ -98,16 +98,14 @@ void FreeFlyerNodelet::Setup(ros::NodeHandle & nh, ros::NodeHandle & nh_mt, std:
   // Immediately, setup a publisher for faults coming from this node
   // Topic needs to be latched for initialization faults
```
bcoltin, line 99: fix comment

bcoltin (approved):
Fix the one comment that's out of date and then looks good.

shared/ff_util/src/ff_nodelet/ff_nodelet.cc
kbrowne15: fixed

PR #675, opened 2023-02-09
support multiple versions for docs

trey0:
Instead of generating documentation only for the `develop` branch, generate and push documentation to GitHub Pages for multiple branches (`develop`, `master`, `ros2`) and their tags. The generated web pages are modified to include a drop-down menu allowing you to switch versions. The default version is `develop` as before.

See deployed example: https://trey0.github.io/astrobee/

Maintenance notes:
- A new version in the menu is added/updated when one of the branches in the list is pushed.
- If the commit you push is a tagged release, a new version is created for the tag (this is a copy and it will persist when the branch is updated later). To ensure you generate docs for a tag, make sure you push the branch and the tag in the same push call, or else you'll have to run the docs workflow manually after you push the tag.
- There is currently no mechanism to go back and generate docs versions for older tagged releases. (But not sure anyone needs that.)
- To prune obsolete tagged releases out of the version list, you can manually check out the `gh-pages` branch, `git rm -r` the version folder, commit and push. (The menu options are auto-detected using the version folders.)

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -4,7 +4,7 @@ name: Build and Push Documentation to gh-pages Branch
 
 on:
   push:
-    branches: ['develop']
+    branches: ['develop', 'master', 'ros2']
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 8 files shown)


marinagmoreira (approved):
Really cool feature!!

PR #676, opened 2023-02-09
Ros2 latched topics

marinagmoreira:


bcoltin (commented):
I think it would be more intuitive if instead of FF_CREATE_PUBLISHER(pub, ...) we did pub = FF_CREATE_PUBLISHER(...). What do you think?

shared/ff_util/include/ff_util/ff_names.h
```diff
@@ -452,4 +455,38 @@
 
 #define SERVICE_STREAMING_LIGHTS                    "hw/signal_lights/streaming"
 
+inline rclcpp::QoS qosType(std::string const& topic, size_t history_depth) {
```
bcoltin, line 458: did we ever include rclcpp in this file? might be relying on header ordering. I wouldn't add it here, I think it would be better to declare a prototype and declare the function elsewhere, just because of how commonly this is included.

bcoltin: Thinking about this a bit more.... it might make sense of we defined the unordered set as a global variable in this file (to have all the topic stuff in one place), then define qosType in the file where we declare the publisher macros?

kbrowne15: Yeah, the problem with that is you can't really define a global variable in a header file because it results in multiple definitions error. I'll work on something else.

bcoltin (approved):
I like it! Thanks!

PR #677, opened 2023-02-09
ROS2 Fam and Ctl

bcoltin:
This compiles and runs but has not been tested yet because it needs inertia and flight mode from the choreographer.

Additionally:
* Removed all the warnings and added Werror
* Changed the timer API slightly @kbrowne15 
* Fixed a bug in ff_flight
* Deleted sim.launch because I kept running the wrong one


gnc/fam/src/fam.cc
```diff
@@ -19,51 +19,52 @@
...
   // config_.AddFile("geometry.config");
   // ReadParams();
   // config_timer_ = nh->createTimer(ros::Duration(1), [this](ros::TimerEvent e) {
   //     config_.CheckFilesUpdated(std::bind(&Fam::ReadParams, this));}, false, true);
 
-  pt_fam_.Initialize("fam");
+  clock_ = nh->get_clock();
 
-  pmc_pub_ = nh->advertise<ff_hw_msgs::PmcCommand>(TOPIC_HARDWARE_PMC_COMMAND, 1);
+  pmc_pub_ = nh->create_publisher<ff_hw_msgs::msg::PmcCommand>(TOPIC_HARDWARE_PMC_COMMAND, 1);
```
kbrowne15, line 41: See comment for ctl.

gnc/fam/src/fam_component.cc
```diff
@@ -19,24 +19,20 @@
...
 #include <memory>
 
 namespace fam {
 
-class FamNodelet : public ff_util::FreeFlyerNodelet {
+class FamComponent : public ff_util::FreeFlyerComponent {
  public:
-  FamNodelet() : ff_util::FreeFlyerNodelet(NODE_FAM, true) {}
-  ~FamNodelet() {}
+  explicit FamComponent(const rclcpp::NodeOptions & options) : ff_util::FreeFlyerComponent(options, NODE_CTL) {}
```
kbrowne15, line 30: Is this supposed to be NODE_CTL or NODE_FAM?

kbrowne15 (approved):

gnc/fam/src/fam_component.cc
bcoltin: good catch!

PR #680, opened 2023-02-15
Make coverage_analyzer.py support Python 3

trey0:
Usually when implementing Python 3 support for a script, you test the Python 2 version under Python 3, get some silly compile errors, fix them, and then the fixed version has identical output under both versions.

In this case, I never got any compile errors under Python 3 (great!) but the results subtly disagreed between the two versions for reasons that I wasn't able to track down (awful).

So what I ended up doing was a fairly substantial refactoring of the code to simplify it in ways that make the behavior easier to understand and predict. As a bonus, the new version runs about 100x faster! In the single real-world test case I got from @amoravargas, it produces the same output as the old version under both Python 2 and 3.

@amoravargas please review (I would have tagged you as a formal reviewer but you were not in the list of options).

### DIFF
--- a/scripts/postprocessing/coverage_analysis/coverage_analyzer.py
+++ b/scripts/postprocessing/coverage_analysis/coverage_analyzer.py
@@ -11,10 +11,9 @@ import subprocess
 import sys
 import timeit
 
+import constants
 import numpy as np
 from tf.transformations import *
-
-import constants
 
 
 class Coverage_Analyzer:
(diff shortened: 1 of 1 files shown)


bcoltin:
Looks fine to me but will wait for feedback from Andres

amoravargas:
Looks good. Thanks for the improvements.

bcoltin (approved):

PR #688, opened 2023-02-21
ros2 upgrade of behaviors/light_flow

ana-GT:
Relevant notes:
* Added **ff_common** export dependency in ff_util. Otherwise, when another package uses ff_util, there are some linking errors.
* Changed from hardware/CMakeLists.txt to hardware/CmakeLists.txt_LEGACY temporarily. This CMake adds all the hardware sub-packages with **add_subdirectory**, which seems to not work well with ament.


bcoltin (approved):

bcoltin:
Thanks Ana!

PR #689, opened 2023-02-22
Improve is_camera robustness, primarily thread safety fixes

trey0:
Summary of changes:
- Removed `ros::spinOnce()` call at the end of `PublishLoop()`. There is a comment to explain why. I consider this change experimental and could probably be convinced to revert it... perhaps there's something I'm not understanding about the ROS nodelet threading model.
- Switched from using ROS timers (which run callbacks in a thread from the nodelet manager's thread pool) to `PeriodicTrigger` instances (which are used to invoke functions in the same thread as `PublishLoop()`). This avoids potential thread safety issues. V4L2 is supposed to be thread-safe but the community wisdom seems to be that it is vendor-specific and you shouldn't count on it. Beyond strict thread safety, a more subtle point is that synchronizing events to happen at a particular point in the frame loop is likely to make the behavior more reliable by avoiding rare corner cases. (For example, what happens if the syscall to adjust the exposure is executed while the camera is in the middle of grabbing a frame? We can eliminate cases like these by syncing the syscall to only happen at a particular point in the loop.)
- Changed newer `ioctl()` calls to use `xioctl()` wrapper, consistent with the rest of the driver.
- Added guards that skip the exposure parameter adjustment syscalls if the actual value was not modified. It's great to avoid syscalls when possible because they typically require a context switch, which adds some latency. (There was also a comment that suggested making too many exposure adjustment calls reduces the frame rate, so it may have more impact than just a context switch.)
- Fixed a bug in modulo arithmetic. The C++ modulo operator preserves the sign of its first operand. You might think that `(i - 1) % k` always evaluates to a number between 0 and k - 1, suitable for indexing into a length-k ring buffer, but in fact it evaluates to -1 when i is 0 and causes an invalid memory reference. The formula should be `(i + k - 1) % k`.

This PR is marked as draft because it should definitely be tested on the granite table before being merged into `develop`. The testing should include applying manual exposure adjustments like we have done in past ISAAC procedures. I don't think there is a huge rush on this. There may be other changes we want to make and test together.

### DIFF
--- a/hardware/is_camera/include/is_camera/camera.h
+++ b/hardware/is_camera/include/is_camera/camera.h
@@ -46,11 +46,11 @@ struct V4LStruct;
 
 class PeriodicTrigger {
  public:
-  inline PeriodicTrigger(unsigned int period) : period_(period), counter_(0) {}
+  inline explicit PeriodicTrigger(unsigned int period) : period_(period), counter_(0) {}
   inline bool IsTriggered() { return counter_ == 0; }
   inline bool Tick() {
     bool result = IsTriggered();
-    counter_ = (counter_ + 1 ) % period_;
+    counter_ = (counter_ + 1) % period_;
     return result;
   }
   inline void SetPeriod(unsigned int period) {
--- a/hardware/is_camera/include/is_camera/camera.h
+++ b/hardware/is_camera/include/is_camera/camera.h
@@ -44,9 +44,29 @@ static void xioctl(int fh, int request, void *arg);
 // Class to hide V4L structures
 struct V4LStruct;
 
+class PeriodicTrigger {
+ public:
+  inline PeriodicTrigger(unsigned int period) : period_(period), counter_(0) {}
+  inline bool IsTriggered() { return counter_ == 0; }
+  inline bool Tick() {
+    bool result = IsTriggered();
+    counter_ = (counter_ + 1 ) % period_;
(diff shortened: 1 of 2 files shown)


bcoltin (approved):

marinagmoreira:
I tested this PR. The nav_cam crashes consistently (and then the entire nodelet group follows) when auto_exposure is set to true in the cameras.config file since startup (not if set to false and then true after starting). (consistently= crashed 3/4 trials)
As a sanity check to make sure this is a problem with this PR I tested `develop` in the same circumstances and in the 3 trials, the fsw didn't crash once.
So there's something fishy happening here at startup...

trey0:
> I tested this PR. The nav_cam crashes consistently (and then the entire nodelet group follows) when auto_exposure is set to true in the cameras.config file since startup (not if set to false and then true after starting). (consistently= crashed 3/4 trials) As a sanity check to make sure this is a problem with this PR I tested `develop` in the same circumstances and in the 3 trials, the fsw didn't crash once. So there's something fishy happening here at startup...

Thanks for checking, Marina! Were you able to get any further information, such as a backtrace from a core dump, or any kind of error message preceding the crash? (If not, I could still try to comb through the logic again to see if there's an obvious mistake.)

trey0:
This PR is broken, closing and capturing the intent in #722

PR #690, opened 2023-02-25
Convert choreographer to ROS2.

bcoltin:
It compiles and runs but no idea if it works (still missing some other parts, like the planner and teleop tool).

It is missing a couple things, but I wanted to get this in since others needed the config and jsonloader stuff.

Things missing:
* Serialization does not work. I think we should convert to json. This will also require converting the existing zone files. Right now I have commented out the zone files.
* Different tests are not converted.

These have both been added to the todo list.

Also, config_client compiles but is untested.


shared/ff_util/include/ff_util/ff_serialization.h
```diff
@@ -36,11 +36,11 @@ class Serialization {
     std::ofstream ofs(file_name, std::ios::out | std::ios::binary);
     if (!ofs.is_open())
       return false;
-    uint32_t serial_size = ros::serialization::serializationLength(msg);
-    boost::shared_array < uint8_t > obuffer(new uint8_t[serial_size]);
-    ros::serialization::OStream ostream(obuffer.get(), serial_size);
-    ros::serialization::serialize(ostream, msg);
-    ofs.write(reinterpret_cast < char* > (obuffer.get()), serial_size);
+    auto serializer = rclcpp::Serialization<RosMessage>();
```
bcoltin, line 39: this probably doesn't work but I think we should replace with json regardless.

shared/ff_util/include/ff_util/ff_service.h
```diff
@@ -100,18 +91,17 @@ class FreeFlyerServiceClient {
...
-  bool Call(FreeFlyerService<ServiceRequestSpec, ServiceResponseSpec> &
-                                                                    service) {
-    return call(service);
+  bool Call(const typename ServiceSpec::Request & request, std::shared_ptr<typename ServiceSpec::Response> & response) {
+    return call(request, response);
   }
 
   // ROS1 functions
-  bool call(FreeFlyerService<ServiceRequestSpec, ServiceResponseSpec>& service) {
+  bool call(const typename ServiceSpec::Request & request, std::shared_ptr<typename ServiceSpec::Response> & response) {
```
bcoltin, line 99: @kbrowne15 I thought it was easier this way. Let me know if you disagree

kbrowne15: Um, so I wonder if we could overload it and do it both ways. The executive, which has 24 service clients, used a service object instead of a request and a response so I was trying to minimize the code changes there. The behaviors also seem to use the service object.

shared/jsonloader/CMakeLists.txt
```diff
@@ -63,69 +53,76 @@ add_library(jsonloader
...
-  )
-  target_link_libraries(test_validation
-    jsonloader
-  )
-
-endif()
+ament_export_targets(jsonloader HAS_LIBRARY_TARGET)
+target_link_libraries(jsonloader gflags glog ${JSONCPP_LIBRARIES} )
+
+#if(CATKIN_ENABLE_TESTING)
```
kbrowne15, line 59: We should make an issue for these tests.

shared/ff_util/include/ff_util/ff_service.h
bcoltin: oh, I didn't know this was used elsewhere. yeah of course.

mobility/choreographer/CMakeLists.txt
```diff
@@ -57,94 +47,89 @@ catkin_package(
...
-      ${catkin_LIBRARIES}
-    )
-  endif()
-endif()
+target_compile_definitions(choreographer
+  PRIVATE "COMPOSITION_BUILDING_DLL")
+ament_target_dependencies(choreographer rclcpp rclcpp_components ff_common ff_util ff_msgs ff_hw_msgs jsonloader geometry_msgs nav_msgs visualization_msgs)
+rclcpp_components_register_nodes(choreographer "choreographer::ChoreographerComponent")
+
+#if(CATKIN_ENABLE_TESTING)
```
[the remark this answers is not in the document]
bcoltin: I added it to the spreadsheet of todo's for ros2, not sure we need an issue as well. (?)

mobility/choreographer/src/choreographer_component.cc
```diff
@@ -1123,138 +1115,128 @@ class ChoreographerNodelet : public ff_util::FreeFlyerNodelet {
...
         // If there is no tolerance violation, reset time
-        tolerance_omega_timer = ros::Time::now();
+        tolerance_omega_timer = GetTimeNow();
       }
     // Send progress in stopping/idling/replanning
     case STATE::STOPPING:
     case STATE::IDLING:
     case STATE::REPLANNING:
-      feedback_.progress = *feedback;
+      // feedback_->progress = *feedback;  // TODO(@mgouveia): ROS2 does not allow this
```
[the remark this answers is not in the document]
bcoltin: I don't think we need to, we'll realize very quickly when we convert whatever uses it :)

mobility/choreographer/CMakeLists.txt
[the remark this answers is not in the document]
bcoltin: I just went ahead and made one, why not

shared/ff_util/include/ff_util/ff_service.h
```diff
@@ -29,21 +30,21 @@
 
 namespace ff_util {
 
-template < class ServiceRequestSpec, class ServiceResponseSpec >
+template < class ServiceSpec >
```
kbrowne15, line 33: Thanks for making this cleaner.

kbrowne15 (approved):

PR #691, opened 2023-03-01
Converted access control to ros2

kbrowne15:


management/access_control/src/access_control.cc
```diff
@@ -91,102 +99,105 @@ std::string AccessControl::GenerateCookie() {
...
+  PublishAck(cmd.cmd_id);
   PublishState();
 }
 
 void AccessControl::PublishAck(std::string const& cmd_id,
                                std::string const& message,
                                uint8_t completed_status,
                                uint8_t status) {
-  ack_.header.stamp = ros::Time::now();
+  ack_.header.stamp = node_->get_clock()->now();
```
bcoltin, line 182: Use the function you made?

bcoltin (approved):

management/access_control/src/access_control.cc
kbrowne15: Good catch. Marina actually made that function.

Issue #628, opened 2023-01-10
Figure out how to migrate ROS1 latching topics to ROS2


trey0:
Migrating ROS1 latching topics to ROS2 was raised as an issue at the FSW meeting on 1/9. The concern is that there is no ROS2 feature that exactly corresponds to ROS1 latching topics, and there are various possible approaches to achieve similar functionality. Creating this thread as one place to discuss.

To over-simplify a bit, the basic issue is that the closest ROS2 feature corresponding to ROS1 latching topics is using "transient local" DDS QoS. However, whereas in ROS1 you only need to declare that a topic is latching on the publisher side, in ROS2 you need to tell both sides to use transient local QoS. (This summary glosses over some other important issues!)

Here are some great background references about the topic:
- https://github.com/ros2/ros2/issues/464 @wjwwood explains why this is a thorny issue. In short, DDS provides transient local QoS which is vaguely similar to ROS1 latching topics, but also has important differences, and there are good reasons to use DDS-native features rather than craft complex ROS2 abstractions on top of them to more precisely mimic older ROS1 functionality.
- https://github.com/ros2/rosbag2/issues/601 One possible approach briefly discussed during the FSW meeting is enabling a subscriber to a topic to use an adaptive QoS that matches the QoS profile used by the publisher. This thread points to three existing implementations of adaptive QoS for subscribers and discusses feasibility of making that functionality more broadly available in ROS2.
- https://github.com/ros2/rclcpp/issues/1868 This thread suggests that adaptive QoS for subscribers will eventually be a core ROS2 feature, but since it hasn't been completed yet, it's unlikely to become available in the ROS2 Humble distro we expect to use for our initial Astrobee ISS deployment of ROS2. In other words, it doesn't make sense to wait for the core feature to be available before tackling this issue.

bcoltin:
From my understanding, much of the reason we need latching topics in our code is because the startup order is non-deterministic and a node will publish an initial message before its subscribers are loaded. I know nothing about the ros2 lifecycle control, but can we use that to avoid the need for this use case?

I'd also ask how many latching topics we have (and how many of them actually need to be latching) which could affect how much effort we want to put into this vs. just using timers and repeating.

trey0:
Another solution approach motivated by the discussion threads above:
- Continue using latching topics (migrating to DDS QoS)
- Use adaptive QoS for subscribers to choose their QoS to match the publishers

[...]

PR #695, opened 2023-03-07
Ros2 migration behavior/arm

ana-GT:
Conversion was straightforward, with the exception of a few issues that I marked TODO:

1. arm/tools/arm_tool.cc, line 168: The arm_tool is initialized using the AnonymousName option, which is not yet available in ROS2, although there seems to be active work on it (as of this week).
2. arm/src/arm_nodelet.cc, lines 491-493: Some functions were not ported, as they no longer seem applicable in ROS2, such as GetPrivateHandle(), Listen function for reconfiguring parameters and its corresponding reconfigure_callback function.
3. arm/src/arm_nodelet.cc: Line 546: A subscriber for TOPIC_JOINT_GOALS is created, originally with latch=true, however  this topic is not in the LatchedTopic list, so if we use the FF_CREATE_SUBSCRIBER, the topic will not be latched.


trey0:
> 1. arm/tools/arm_tool.cc, line 168: The arm_tool is initialized using the AnonymousName option, which is not yet available in ROS2, although there seems to be active work on it (as of this week).

As far as I understand, the `AnonymousName` option appends a random number to the end of the node name at initialization time, which we could do manually at [1]. I see you also changed the node name from `"control"` to `"arm_tool"`, which seems like an improvement.

[1] https://github.com/nasa/astrobee/blob/939adfe9cb50e536800afb6937c21793a64e3e41/behaviors/arm/tools/arm_tool.cc#L195

trey0:
> 3. arm/src/arm_nodelet.cc: Line 546: A subscriber for TOPIC_JOINT_GOALS is created, originally with latch=true, however  this topic is not in the LatchedTopic list, so if we use the FF_CREATE_SUBSCRIBER, the topic will not be latched.

@marinagmoreira and @kbrowne15 recently merged https://github.com/nasa/astrobee/commit/a3ae21283d311fcc9c48d3cf75fb4d76bc41ed07 on `develop`, which makes this publisher non-latching.

I'm not sure why it was originally marked latching. The latching option would have the effect of automatically resending the joint goals if somehow the arm nodelet joined late after the goals were originally published, but I don't think that's really desired (could lead to an unintended "zombie command").

marinagmoreira:
So I tested the arm behavior in simulation and it works. But let's hold on merging this until we make a decision on the services since it affects it.
I commented out the services test because as a gtest, it's really hard to test it since the server and client are started from the same main, and that opens a can of worms for the sync service problem. That's why the service test passed in the past, even if we couldn't use it for components.
I'll keep on tinkering with the service test to see if I can make it work with a multithreaded executor since we might want to have the server and client under the same container and if this doesn't work, that's just problems down the line.. just don't want to sink even more time into it :(..

behaviors/arm/src/arm_nodelet.cc
```diff
@@ -507,83 +536,78 @@ class ArmNodelet : public ff_util::FreeFlyerNodelet {
...
+        std::bind(&ArmComponent::TimeoutCallback, this), nh, true, false);
 
     // Publishers for arm and joint state
-    sub_joint_states_ = nh->subscribe(TOPIC_JOINT_STATES, 1,
-      &ArmNodelet::JointStateCallback, this);
-    pub_joint_goals_ = nh->advertise<sensor_msgs::JointState>(
-      TOPIC_JOINT_GOALS, 1);
+    sub_joint_states_ = FF_CREATE_SUBSCRIBER(nh, sensor_msgs::JointState, TOPIC_JOINT_STATES, 1,
+      std::bind(&ArmComponent::JointStateCallback, this, std::placeholders::_1));
+    // TODO(ana): This had a latch=true argument, but TOPIC_JOINT_GOALS is NOT in the LatchedTopic list
```
marinagmoreira, line 549: According to the analysis in https://github.com/nasa/astrobee/issues/628, this does not need to be latched

behaviors/arm/src/arm_component.cc
```diff
@@ -17,43 +17,66 @@
...
-  };
+enum ServoID {
+  PROXIMAL_SERVO,           // Proximal joint servo
+  DISTAL_SERVO,             // Distal joint servo
+  GRIPPER_SERVO,            // Gripper joint servo
+  ALL_SERVOS          // Proximal, Distal and Gripper servos
+};
 
 // Joint information, where HUMAN = SCALE * DRIVER + OFFSET
 struct JointInfo {
```
bcoltin, line 82: Now that we aren't using tol (I'm pretty sure) except to set it but never look at it, I would just delete it from JointInfo to make things more clear.

bcoltin (approved):
Thanks Ana and Marina!

behaviors/arm/src/arm_component.cc
trey0: That sounds like a change to apply/test on `develop` first?

bcoltin: I thought it only applies here due to changes in how the parameter server works, but correct me if I'm wrong Marina.

marinagmoreira: In order to maintain functionality with the previous version the tol with GetTolerance() still only gets updated in certain flight modes, which maintains the need for the tol variable :)

PR #702, opened 2023-03-15
vocab db: report memory usage

trey0:
Added methods to report approximate memory usage of protected class members in the bag of words data structures. The methods are designed to be called from a higher-level `ReportMemoryUsage()` function in the `sparse_mapping` package.

This PR needs to be merged into `master` and deployed to `astrobee_base` before the higher-level code will pass tests.

The larger aim is to figure out the main contributors to sparse map memory usage so we can evaluate possible optimizations.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -49,7 +49,7 @@ Depends: ${misc:Depends}
   ${ros-python},
   astrobee-config (>= ${binary:Version}),
   astrobee-comms (>= ${binary:Version}),
-  libalvar2 (>=2.0), libdbow21 (>=0.1), libgtsam (>=4.0), libopenmvg1 (>=1.0), libroyale1 (>=1.0),
+  libalvar2 (>=2.0), libdbow21 (>=0.1-6), libgtsam (>=4.0), libopenmvg1 (>=1.0), libroyale1 (>=1.0),
   libceres1 (>=1.0), rti (>=1.0), libmiro0 (>=0.1), libsoracore1 (>=1.0),
   libdecomputil0 (>=0.1), libjps3d0 (>=0.1),
   libluajit-5.1-2,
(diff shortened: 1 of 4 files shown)


bcoltin:
If we update this, it is going to take some extra effort from the facility to track debian versions for dbow2. Which is fine, they can do that, but I am not sure if it is worth it for a debugging change we can run on the ground with the same map file and get the same results. If we do need to check the size regularly, we could get a rough estimate from the public size function. What do you think, is this needed long-term, or would running it just when needed be sufficient?

trey0:
> If we update this, it is going to take some extra effort from the facility to track debian versions for dbow2. Which is fine, they can do that, but I am not sure if it is worth it for a debugging change we can run on the ground with the same map file and get the same results. If we do need to check the size regularly, we could get a rough estimate from the public size function. What do you think, is this needed long-term, or would running it just when needed be sufficient?

Those are great points.

It would be totally valid to have this capability available only on the ground, but I think our current tooling makes it difficult to provide features on the ground while not deploying them on-orbit.

There are public `size()` functions, but they return a count of elements rather than an estimate of memory usage.

I would like to have the ability to run this query again on future maps or with future changes to our sparse map representation, without forcing people through a complex process of switching branches and rebuilding dbow2, so I think it's worth cutting a new debian to allow that.

Coming out of this discussion, some new tasks before this should be merged:
- [ ] Update the patchlevel of the debian
- [ ] Update the changelog
- [ ] Update any `dbow2` dependencies in other debian control files to require the new patchlevel

bcoltin:
Also: 

- [ ] return needed items publicly from dbow, move prints into our function (so we can change it without updating the debian)

Nesha2023 (approved):

bcoltin (approved):
Thanks!

PR #707, opened 2023-03-23
Ros2 migration of localization_manager package

ana-GT:
I converted this package since it appeared as  a dependency of the behaviors/perch package (in closer inspection, that dependency wasn't really needed :) ).

Conversion was straightforward, tested running Astrobee in Gazebo and seeing that the services/action are available. Only couple of notes:

1. When a localization pipeline is shutdown, the related subscribers called the **shutdown()** function. For ROS2, there is not a direct equivalent that I am aware of, so I am using **.reset()** , as the subscribers are shared pointers. Just mentioning this in case this replacement is not desired.
3. **config_server_** no longer has the **Lim** function, which appear in 2 instances in this package. I just deleted these instances, in a similar manner as how it was done in the choreographer ros2 update.


marinagmoreira:
> I converted this package since it appeared as a dependency of the behaviors/perch package (in closer inspection, that dependency wasn't really needed :) ).
> 
> Conversion was straightforward, tested running Astrobee in Gazebo and seeing that the services/action are available. Only couple of notes:
> 
> 1. When a localization pipeline is shutdown, the related subscribers called the **shutdown()** function. For ROS2, there is not a direct equivalent that I am aware of, so I am using **.reset()** , as the subscribers are shared pointers. Just mentioning this in case this replacement is not desired.
> 2. **config_server_** no longer has the **Lim** function, which appear in 2 instances in this package. I just deleted these instances, in a similar manner as how it was done in the choreographer ros2 update.

yup, .reset() is how I'm converting the shutdown() too :)

marinagmoreira (approved):

PR #710, opened 2023-03-28
Import certificate for Astrobee server SSL support

rgarciaruiz:
* Add certificate import to install process
* Minor doc update
* Sync platform submodule pointer


bcoltin (approved):
Thanks!

PR #729, opened 2023-05-02
localization_analysis remove extra flags + docs

marinagmoreira:
- simplifying localization_analysis arguments based on env variables that need to be set for sparse_mapping anyway
- adding script output
- Make testing against a localization map optional
- some docs

### DIFF
--- a/localization/localization_common/scripts/localization_common/utilities.py
+++ b/localization/localization_common/scripts/localization_common/utilities.py
@@ -74,13 +74,26 @@ def load_dataframe(files):
 
 
 def run_command_and_save_output(command, output_filename="", print_command=True):
+    """
+    Run a command in the shell, save the output to a file, and print the output.
+
+    Parameters:
+        command (str): The shell command to be executed.
+        output_filename (str): (optional) The filename to save the command output. Default is an empty string.
+        print_command (bool): (optional) If True, print the command before executing. Default is True.
+
+    Returns:
+        tuple: A tuple containing the return code, standard output, and standard error.
+    """
 
     if print_command:
         print(command)
 
+    # If an output_filename is provided, open the file for writing
     if output_filename != "":
         f = open(output_filename, "w")
 
+    # Run the command in a subprocess and capture the output
     stdout = ""
     stderr = ""
     popen = subprocess.Popen(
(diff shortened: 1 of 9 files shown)


localization/localization_common/scripts/localization_common/utilities.py
```diff
@@ -73,11 +73,40 @@ def load_dataframe(files):
     return dataframe
 
 
-def run_command_and_save_output(command, output_filename, print_command=True):
+def run_command_and_save_output(command, output_filename="", print_command=True):
```
rsoussan, line 76: Nice! Does this now alert the user when the script fails? Can you add some comments to the code blocks?

tools/localization_analysis/scripts/make_map.py
```diff
@@ -104,10 +91,35 @@ def make_map(
 
         # Link maps directory since conversion to BRISK map needs
         # image files to appear to be in correct relative path
-        os.symlink(maps_directory, "maps")
-        maps_bag_images = os.path.join("maps", bag_images_dir)
-        if not os.path.isdir(maps_bag_images):
-            os.symlink(bag_images, maps_bag_images)
+        build_map_command = (
```
rsoussan, line 94: Can you add some details here? Was the symlink messing up sometimes?

tools/localization_analysis/scripts/make_groundtruth.py
```diff
@@ -41,24 +41,19 @@
         help="Existing map to use as basis for groundtruth.  Should largely overlap area covered in input bagfile.",
     )
     parser.add_argument(
-        "maps_directory",
-        help="Location of images used for each bagfile use to generate base_surf_map.",
+        "--loc-map",
+        default="",
+        help="Full path to Localization map for bagfile to test localization performance.",
```
rsoussan, line 46: Add that if no loc map is passed, the localization test is not run

tools/localization_analysis/scripts/make_map.py
marinagmoreira: I'm really pulling my memory here, but if I remember correctly the problem was that if you symlink to `maps`, you're changing the dir name, so the relative path didn't work?

We're still doing very similar symlinks, in line 117 forward, the difference is that I'm getting the path location from the map itself instead of having the user specify it manually (using the build_map -info command)

localization/localization_common/scripts/localization_common/utilities.py
marinagmoreira: Added comments to the code. Yes it alerts the user by printing the error message and exits the program

tools/localization_analysis/scripts/make_groundtruth.py
marinagmoreira: added that comment

PR #733, opened 2023-05-17
ROS2 mobility: (part of) planner_qp 

joris997:
this is a PR for the planner_qp package for the ROS2 conversion. Deliberately, it does not include planner_qp itself, as I first wanted to check this in and gather any feedback.

For reference, this PR therefore includes
- traj_opt_basic
- traj_opt_msgs
- traj_opt_pro
- traj_opt_ros

@marinagmoreira 

In traj_opt_ros, I separated the nh_ and the pub map, into one for Trajectory and SolverInfo, as it seems ROS2 does not allow the implementation that ROS1 allows. The code does seem to allow this change, based on the `getPub`, `getInfoPub`, and the other methods were already explicit.

Cheers,
Joris


Pedro-Roque:
@marinagmoreira not sure if you are going to be involved in the ROS2 port, but I realized that this PR was still waiting feedback.

PR #736, opened 2023-05-31
adding CI that compiles jar messages for gs utilization

marinagmoreira:
this is mostly copied over from existing isaac functionality

also:
-merged push and release actions into one and disabled image pushing outside of the nasa account to minimize memory usage for forks


.github/workflows/msgs_jar.yaml
rgarciaruiz, line 1: Looks great! 

So the idea is that users would pull the image and repeat the docker cp part? Or is this available on GitHub in some way I'm missing?

marinagmoreira: after this CI runs for the first time after a push, you can do from any machine with docker installed:
`docker cp $(docker create --rm ghcr.io/nasa/astrobee:latest-msgs-jar):/src/msgs/devel/share/maven/ .`
anf the folder with the .jar files will be copied. Images get uploaded here: https://github.com/nasa/astrobee/pkgs/container/astrobee

rgarciaruiz: I know, what I mean is that there is a `docker pull` implied in the process. Not a huge deal since the image is not that big. But I still wonder if we should use `upload-artifact` or something similar to make the jars directly available without the users having to install docker and pull the image. I haven't used that action myself but I believe it's meant for cases like this.

marinagmoreira: that's definitely better! https://github.com/marinagmoreira/astrobee/actions/runs/5183166184

rgarciaruiz: Great! 

Now, do you think this is accessible enough to users? How should we direct them into finding this artifact? Or do you think we should make this some sort of independent workflow separate from PRs so we don't have that many versions of the files? Maybe make it similar to the docs workflow?

Also, I'm not sure if this is a problem, but we should probably make the artifacts unique:
image
https://github.com/actions/upload-artifact

.github/workflows/ci_push.yml
```diff
@@ -20,9 +22,27 @@ jobs:
...
     - name: Build, test, push docker
-      run: ./scripts/docker/build.sh --xenial --remote
-                                     --owner ${{ github.repository_owner }}
-                                     astrobee test_astrobee push_astrobee
+      run: >
+        revision="latest";
+        build_arg="astrobee";
+        push_arg="";
+        branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}};
+        if [ $branch == "nasa" ]; then
```
rgarciaruiz, line 30: Branch nasa? This job is different from the others, was that intentional?

marinagmoreira: this is a bug hahah thanks for spotting it!

.github/workflows/msgs_jar.yaml
marinagmoreira: I don't think that's a problem because that action only has one job.

How we should direct the users is a really good question because the paths become old when new commits are added.  The easiest way I can think of is to create a status badge that links to the action run list and ask them to click on the latest like:

[Build jar messages](https://github.com/marinagmoreira/astrobee/actions/workflows/msgs_jar.yaml)

rgarciaruiz: That looks good. Yeah, good idea. 

That includes PR versions though, do you see a problem with that? Should we restrict it to push/develop or do you see a value in having it as a test for PRs? I'm just worried that people would use an "unofficial" jar file. I guess we could also tell them to filter by actor or branch (maybe even add that to the badge)

Let me know your thoughts.

marinagmoreira: Good point, let's filter by branch and make the status to update on push:
branch master:
[Build jar messages](https://github.com/nasa/astrobee/actions/workflows/msgs_jar.yaml?query=branch%3Amaster)
or 
branch develop:
[Build jar messages](https://github.com/nasa/astrobee/actions/workflows/msgs_jar.yaml?query=branch%3Adevelop)

.github/workflows/ci_push.yml
```diff
@@ -20,9 +22,27 @@ jobs:
...
-      run: ./scripts/docker/build.sh --xenial --remote
-                                     --owner ${{ github.repository_owner }}
-                                     astrobee test_astrobee push_astrobee
+      run: >
+        revision="latest";
+        build_arg="astrobee";
+        push_arg="";
+        branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}};
+        if [ "${{ github.repository_owner }}" == "nasa" ]; then
+          if [ "${{ github.head_ref }}" == "develop" ]; then
```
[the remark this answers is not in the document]
marinagmoreira: that is correct, and if that doesn't exist it uses GITHUB_REF that is available on push and does some editing to just end up with the branch name. I did confirm that this worked by printing it on the PR and local push action on my fork :)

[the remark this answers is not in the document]
rgarciaruiz: Ok, good, but I meant line 31, would that fallback too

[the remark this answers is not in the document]
marinagmoreira: oh yeah I need to replace it with the $branch var

.github/workflows/msgs_jar.yaml
rgarciaruiz: Where is this status going then? astrobee_android? somewhere on the readme here? both?

marinagmoreira: I'm assuming astrobee_android apk install instructions, not sure where to put it on an astrobee readme.

Also thinking about it, just creating the artifact on pushes to develop and master on the NASA repo seems enough and still keeping the build part for the PR's/others. I'll change that :)

rgarciaruiz (approved):
Thank you for working on this, Marina

PR #737, opened 2023-06-08
fix: use join function instead using write function and using write f…

slowy07:
improvement on ``eps_aux.py``:
- using f string for print format instead using old style python ``%``, and make more concise and readable formatting code, reference: [PEP 498](https://peps.python.org/pep-0498/)

improvement on ``luaTable.py``
- using f string for print format instead using old style python ``%``, and make more concise and readable formatting code, reference: [PEP 498](https://peps.python.org/pep-0498/)
- using ``join`` for concatenating lines instead of repeatedly calling ``write`` for each line, using ``join`` to concatenate the lines and write them in a single call
- adding type hinting to the function parameters, reference [PEP484](https://peps.python.org/pep-0484/)

### DIFF
--- a/hardware/eps_driver/tools/eps_aux.py
+++ b/hardware/eps_driver/tools/eps_aux.py
@@ -11,7 +11,7 @@ def callback(data):
             aux = c.value
         if c.name.find("SYSTEM_I") >= 0:
             system = c.value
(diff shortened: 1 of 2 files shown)


trey0:
Closing spam PR

PR #740, opened 2023-07-18
Draft: Nominal MPC support package

Pedro-Roque:
This PR contains an Astrobee ctl class for Nominal MPC operations. Further integration with the choreographer to support both the onboard PID and the MPC in runtime is needed and should be added.


bcoltin:
If we want to merge this, we'd also need to remove asado from the repo and link to an installed version.

Pedro-Roque:
> If we want to merge this, we'd also need to remove asado from the repo and link to an installed version.

Or should we add acado to external depencies? We can link to system installed versions with no problem either.

bcoltin:
> > If we want to merge this, we'd also need to remove asado from the repo and link to an installed version.
> 
> Or should we add acado to external depencies? We can link to system installed versions with no problem either.

Yeah, I think that's what I meant.

PR #744, opened 2023-07-20
Remote docker container

marinagmoreira:
- Based on https://github.com/nasa/astrobee/pull/622
- Tested devcontainer both on beehive and bartlett and both ssh and vnc worked without any setup (ssh on display :1 and vnc on display :0)
- Also tested that astrobee simulation runs on both ssh and vnc

- [x] Add documentation


trey0 (changes requested):
So glad this is finally working the way you wanted!

We definitely need to fix the `-p` argument security issue in my comment. The other stuff is less important.

marinagmoreira:
given the security concerns of running this with docker directly. I'm closing this PR for now and opening a different one using vscode. The new one offers both the x11 and vnc route with much simpler instructions.

The PORT is still not configurable becausause to my knowledge you can't use variables in the `devcontainer.json` for forwardPorts. We can instead make a script that changes the port on all the files

PR #755, opened 2023-11-16
fix fov calculation, not considering distortion

marinagmoreira:
The previous formula was just not correct. Fixed it to be accurate without distortion.
This can be further refined by using https://github.com/nasa/isaac/blob/master/astrobee/behaviors/inspection/scripts/field_of_view_calculator.py but I figured we fix it first since some isaac code depends on it to run well


trey0 (approved):
Good catch!

PR #756, opened 2023-11-20
Comms bridge

kbrowne15:
This is the comms bridge that allows any ros message to be sent between robots without having to specifically add ros to dds and vice versa code.


trey0 (commented):
Thanks for sharing this! Great to see how far it's coming along.

Comments mostly trivial.

I had a hard time understanding how to estimate the memory usage. The 128K per content message buffer allocation could potentially get bad quickly if we aren't careful. But maybe it doesn't matter for ISAAC in the near term if we don't subscribe to high-rate messages.

astrobee/config/communications/comms_bridge.config
```diff
@@ -0,0 +1,74 @@
...
+--   queue_size_ABREV int:    queue size --
+
+require "context"
+
+-- Enable DDS communications on start --
+-- If false, a trigger call will be needed to start communications --
+started = false;
+
+-- Rapid subscribers --
+rapid_sub_names = {"Bumble", "Honey", "Queen", "Bsharp", "Wannabee"}
```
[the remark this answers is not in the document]
kbrowne15: I'm not entirely sure what you are asking but thank you for asking it as just realize I have a bit of undesired behavior that I need to fix. As you know, the relay code converts a ros message into an advertisement info message (only the first time it sees the message type) and the content message. I was just publishing the messages on dds topics "generic_comms_advertisement_info" and "generic_comms_content". However this means all connected robots will receive all topics so I will change this to append the receive robot's name to it.

astrobee/config/communications/dds_generic_comms/NDDS_DISCOVERY_PEERS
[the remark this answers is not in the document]
kbrowne15: I'm not sure what you mean by placeholder but this file will be populated with the ISS IP addresses of the robots the current robot wants to connect to. A comment can be added and a readme will be written before this becomes an actual pr and not just a draft.

astrobee/config/communications/comms_bridge.config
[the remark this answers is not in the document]
trey0: Ha, glad you caught that.

But I was talking about a different issue: when a ROS message is relayed from robot A to robot B, how do we specify the relationship between the topic used locally on A and the topic of the relayed message that the bridge publishes on B?

In my draft spec for the config file, the idea was that you could either manually specify both the "from" and "to" topics in the relay entry, or if you prefer, you could specify only the "from" topic and it would auto-generate the "to" topic using a default template string.

I think I understand now that your concept is to hard-code the relationship between "from" and "to" topics because we need it to work a certain way for compatibility with the topic naming convention for multi-robot sims. That's probably ok, but I think it should be documented here, something like:

"The topic `$topic` you specify in this config is a suffix of the full topic name and must not start with `/`. It will be combined with different prefixes based on the context. When relaying from `$from_robot` to `$to_robot`, the `from` side of the bridge subscribes to the local topic `/$topic` and the `to` side of the bridge publishes the namespaced global topic `/$from_robot/$topic`. This convention is designed for compatibility with how we run multi-robot sims. It allows us to specify that a ROS node running on `$to_robot` should communicate with ROS nodes on `$from_robot` (instead of local ROS nodes) by launching it with the command-line argument `-ns $from_robot`."

I also have a trivial suggestion that in each of these relay entries, we should call the main field `topic` instead of `name`, just to be self-documenting.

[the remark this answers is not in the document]
kbrowne15: I apologize, I opened your draft spec for the config file in a tab and completely forgot to read it before working on the bridge. I love your idea of having an optional "to" topic and generating the "to" topic by appending the robot name to the beginning of the topic if the "to" topic isn't specified. I plan to add that this week and I also plan to thoroughly document this before putting the pr in official review. Further, I will change name to topic. :)

astrobee/config/communications/comms_bridge.config
```diff
@@ -0,0 +1,53 @@
...
+-- License for the specific language governing permissions and limitations
+-- under the License.
+
+require "context"
+
+-- Enable DDS communications on start --
+-- If false, a trigger call will be needed to start communications --
+initialize_dds_on_start = false
+
+links = {
```
marinagmoreira, line 24: could we have 2 links, one for granite, the other for iss? That way we don't have to edit this to test on the granite :)

communications/comms_bridge/src/bridge_publisher.cpp
```diff
@@ -0,0 +1,240 @@
...
+  ros::serialization::IStream stream(const_cast<unsigned char*>(content_info.data.data()), content_info.data.size());
+  try {
+    stream.next(shifter);
+  }
+  catch (ros::Exception &e) {
+    ROS_ERROR("Deserialization error (%s), losing message", e.what());
+    return false;
+  }
+
+  // will always succeed, even if inconsistent type info
```
bcoltin, line 121: Should we at least print an error or something if this happens?

scripts/git/cpplint.py
```diff
@@ -6660,8 +6660,8 @@ def FlagCxx11Features(filename, clean_lines, linenum, error):
     if include and include.group(1) in (
         "cfenv",
         "fenv.h",
-        "condition_variable",
-        "future",
+        # "condition_variable",
```
bcoltin, line 6663: why?

marinagmoreira: the linter wasn't passing on Michael's code because he had headers "condition_variable" and "future", so I added it to the allowed header list

marinagmoreira (approved):
Tested latest version in the lab

rgarciaruiz (approved):
The current configuration for the bridge is not exactly what other users would need. Usually, they need /gnc/ekf both ways. This is not a big deal though since it can be configured.

bcoltin (approved):

astrobee/config/communications/comms_bridge.config
```diff
@@ -0,0 +1,77 @@
...
+    relay_forward = {
+    },
+    relay_backward = {
+    },
+    relay_both = {
+      {in_topic = "gnc/ekf"},
+      {in_topic = "gs/data"},
+    },
+  },
+  {
```
trey0, line 61: I like how this is structured! But it would help to have a comment here explaining that the second section is for granite table use and will be ignored when the bridge is used on ISS because the robot names are different. (I assume that's what's going on.)

astrobee/config/communications/comms_bridge.config
```diff
@@ -0,0 +1,77 @@
...
+    },
+    relay_both = {
+      {in_topic = "gnc/ekf"},
+      {in_topic = "gs/data"},
+    }
+  }
+}
+
+verbose = 2
+ad2pub_delay = 3.0
```
trey0, line 77: It would help to explain what these fields do and how they should be configured. (Or at least reassure the user that they don't commonly need to be edited.)

astrobee/config/communications/dds_generic_comms/NDDS_DISCOVERY_PEERS
```diff
@@ -0,0 +1,17 @@
...
+;builtin.shmem://        ; The builtin builtin 'shmem' transport plugin
+                         ;  at network address FCC0::0
+;
+;; Unicast
+;
+;; Localhost
+127.0.0.1              ; A comma can be used a separator
+;
+;; Robot to communicate with
+;
```
trey0, line 16: I take it this is a placeholder example file that needs to be edited for actual use of the bridge. Could you add an overview comment explaining that?

And the comments that are here could be clarified, like "Robot to communicate with" could be "Add IP address of robot to communicate with here" (assuming that's what's intended).

Also, if there are some sections we don't expect to be used for the bridge (like shared memory), maybe the file could be rearranged so they appear at the end under a section header like "Advanced features (you probably don't need to edit below this point)".

communications/comms_bridge/readme.md
```diff
@@ -0,0 +1,50 @@
...
+## Enabling DDS Communication
+
+The generic comms bridge runs by default when the astrobee flight starts up. By
+default, dds publishing and subscribing is disabled since astrobee to astrobee
+communication is not desired for every activity. There are 2 ways to enable it.
+One can either change the 'initialize_dds_on_start' parameter in the comms bridge
+config file to true before starting flight software or send an "enable astrobee
+intercomms" command to astrobee after the flight software has started.
+
+## Specifing Topics
```
trey0, line 28: Specifying

communications/comms_bridge/readme.md
```diff
@@ -0,0 +1,50 @@
...
+### QoS Profile
+
+By default, DDS properties are used to ensure the advertisement info messages
+are reliably deliveried to the republisher class even if the subscriber or
+republisher are restarted. Currently, DDS is configured to work if there are
+no more than 20 topics being passed between robots. If more topics are required,
+please change the message depth in the "RAPID_QOS_PROFILES.xml" file in the
+"dds_generic_comms" folder. The depth for the "RapidReliableDurableQos" will
+need to be changed for both the "datareader_qos" and the "datawriter_qos". This
+should be on or around lines 206 and 218 in the file.
```
trey0, line 50: Thank you for documenting this issue! Sorry I'm too lazy to figure out if you already did this, but will there be a warning message, or possibly a fatal error, if it looks like there are too many topics? (It seems like it would be easy to check right at initialization based on counting the topics in the config file.)

scripts/git/pre-commit.linter_python
```diff
@@ -71,11 +71,11 @@ echo "  Analysing python code style with 'isort'."
 # could happen for example if the .isort.cfg src_paths list gets out
 # of date.
 
-if $(isort . --extend-skip cmake --profile black --diff --check-only --quiet >/dev/null); then
+if $(isort . --extend-skip cmake,submodules --profile black --diff --check-only --quiet >/dev/null); then
```
trey0, line 74: Interesting. Should we propagate this change to the `isaac` repo? Not sure why the relevant commands are different.

https://github.com/nasa/isaac/blob/develop/scripts/git/pre-commit.linter_python#L75-L79

trey0 (approved):
Thanks! I'm really excited about this coming together and hope other Astrobee users will get some benefit. My comments are minor and mostly about docs.

scripts/git/pre-commit.linter_python
trey0: (To answer my own question: "Yes." https://github.com/nasa/isaac/issues/123 )

communications/comms_bridge/readme.md
kbrowne15: I can definitely count the topics in the config file but there is no easy way to get the queue size out of the qos profiles file. I'll add output for the default value, 20, and make a comment that if it is changes in the qos profile file that it also needs to be changed in the check in the code.

PR #763, opened 2024-02-08
Changed honey nav cam gain and exposure to match bumble.

kbrowne15:


bcoltin:
We didn't intentionally calibrate it this way for some reason?

marinagmoreira:
> We didn't intentionally calibrate it this way for some reason? (just checking)

From what Ruben said, the Bumble values were adjusted from ISS activities a long time ago. But the other robot config files weren't updated to match it, so when Honey was deployed on-orbit, it had the old values

marinagmoreira (approved):

bcoltin (approved):

PR #764, opened 2024-02-09
Prevent CPU Mem Monitor Blocking on Startup

bcoltin:
Starts a separate thread to get all PIDs so the heartbeat can be published immediately.

### DIFF
--- a/astrobee/config/management/cpu_mem_monitor.config
+++ b/astrobee/config/management/cpu_mem_monitor.config
@@ -18,7 +18,6 @@
 -- frequency to check and publish cpu stats
 
 llp = {
-  update_pid_hz = 0.1,
   update_freq_hz = 1,
   temperature_scale = 0.001,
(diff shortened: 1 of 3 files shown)


kbrowne15 (approved):

marinagmoreira (approved):
looks good! was this tested on the robot?

bcoltin:
No, only on the simulator. Should we wait to merge until testing on the robot? Probably makes more sense to wait until next testing session rather than going to lab just for this.

marinagmoreira:
> No, only on the simulator. Should we wait to merge until testing on the robot? Probably makes more sense to wait until next testing session rather than going to lab just for this.

no worries, I'll test it quickly and merge this in :)

bcoltin:
Thanks!

marinagmoreira:
tested in the lab

bcoltin:
Thanks!

PR #770, opened 2024-03-06
ROS Noetic Java Generator

rgarciaruiz:
Debians created from these scripts contain a minimal set of patches that enable generation of Java messages on ROS Noetic.


trey0:
Hi, Ruben- I'm excited to see your progress with this! I added a suggested edit to the system requirements section of the install instructions, assuming I understand what you're trying to do here.

Some other suggestions for useful additions if you have the time:
- Updated documentation about the new way to build the JAR file. (I don't know which repo that goes in or if there is any old README that can be edited, vs. we have to create instructions from scratch.)
- Possibly a new CI action for GitHub that successfully builds the JAR file. That's always good for peace of mind.

Thanks!

rgarciaruiz:
Thank you for the doc update! 

As for the remaining of the documentation, you can take a look at [this PR](https://github.com/nasa/astrobee_android/pull/64). Let me know if you have any comments.

As for the CI action, good idea. I may just repurpose the `build_msgs_jar.Dockerfile` for that.

scripts/docker/build_msgs_jar.Dockerfile
```diff
@@ -21,18 +21,23 @@
...
+RUN apt-get update && apt-get install -y devscripts equivs
+
+# make sure there is a python binary
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 1
+
+# build rosjava debians
+RUN cd /src/msgs/src/scripts && /bin/bash build_rosjava_debians.sh --install-with-deps
+
+# remove to save disk space
+RUN rm -rf /var/lib/apt/lists/*
```
trey0, line 38: Not a big deal, but a point of interest is that if you want to optimize the size of a docker image by eliminating temp files, it's important to combine the temp file generating action with the cleanup action in the same `RUN` step, like `RUN make files && rm files`. This is because `docker build` creates a new image layer for each `RUN` command, so if you separate the actions in different steps, there will still be a layer containing the temp files (and using disk space) even if the temp files are not present in the top layer.

(Or you can take a different approach and squash all the layers into one at the end before export, in which case this becomes a non-issue.)

trey0 (approved):
Looks great! Thank you!

scripts/docker/build_msgs_jar.Dockerfile
rgarciaruiz: Good point, I think I'll move the `rm` to the previous RUN only, since the build_rosjava_debians.sh needs the apt lists from the previous command and I don't think combining them will make much of a difference with size.

PR #773, opened 2024-03-25
Add exposure service

marinagmoreira:
Adding change exposure service, such that we can add it as an executive command

### DIFF
--- a/hardware/is_camera/src/camera.cc
+++ b/hardware/is_camera/src/camera.cc
@@ -258,7 +258,8 @@ namespace is_camera {
     } else {
       // Failed
       res.success = false;
-      res.status_message = "Failed! Either thread not running or auto-exposure on";
+      res.status_message = "Failed! Either thread not running (" + std::to_string(thread_running_) +
+                           ") or auto-exposure on (" + std::to_string(auto_exposure_) + ")";
     }
 
     return true;
--- a/hardware/is_camera/src/camera.cc
+++ b/hardware/is_camera/src/camera.cc
@@ -248,11 +248,19 @@ namespace is_camera {
   // Set the exposure
   bool CameraNodelet::SetExposure(ff_msgs::SetExposure::Request  &req,
                           ff_msgs::SetExposure::Response &res) {
-    // Set exposure
-    camera_exposure_ = req.exposure;
-    // Success!
-    res.success = true;
-    res.status_message = "Success";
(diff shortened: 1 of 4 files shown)


kbrowne15 (approved):

hardware/is_camera/src/camera.cc
```diff
@@ -243,6 +245,25 @@ namespace is_camera {
...
+      // Set exposure
+      camera_exposure_ = req.exposure;
+      v4l_->SetParameters(camera_gain_, camera_exposure_);
+      // Success!
+      res.success = true;
+      res.status_message = "Success";
+    } else {
+      // Failed
+      res.success = false;
+      res.status_message = "Failed! Either thread not running or auto-exposure on";
```
trey0, line 261: It would probably be better to include the values of `thread_running_` and `auto_exposure_` in this message.

hardware/is_camera/src/camera.cc
```diff
@@ -243,6 +245,25 @@ namespace is_camera {
     thread_ = std::thread(&CameraNodelet::PublishLoop, this);
   }
 
+  // Set the exposure
+  bool CameraNodelet::SetExposure(ff_msgs::SetExposure::Request  &req,
+                          ff_msgs::SetExposure::Response &res) {
```
trey0, line 250: I thought I remembered working on something like this before... here it is https://github.com/nasa/astrobee/pull/689 . That PR was rejected because it caused some crashes but we never learned why.

Some things I like about that PR that could make it into this if it makes sense:
- It separates out `SetExposure()` and `SetGain()` functions so they can be called from wherever in a way that would make this PR more convenient. There is also a guard to avoid making a redundant syscall if the desired value is already the current value.
- I noticed some `ioctl()` calls that I guess should be converted to `xioctl()` in general.

I don't feel strongly about including this stuff if it feels too tangential, but it could be nice to get it in if it can piggyback on your testing plans.

trey0 (approved):
Looks good to me! Just a couple of minor comments, no blockers.

hardware/is_camera/src/camera.cc
marinagmoreira: Hi Trey! Since I'm in a rush to merge it such that Katie can add the command to the executive before leaving, I'll delay doing this, but I'm always available for lab testing on any PR :)

PR #775, opened 2024-03-27
Added publishing messages at a certain rate to the comms bridge.

kbrowne15:


astrobee/config/communications/comms_bridge.config
```diff
@@ -83,7 +88,7 @@ links = {
     relay_backward = {
     },
     relay_both = {
-      {in_topic = "gnc/ekf"},
+      {in_topic = "gnc/ekf", rate=1},
```
rgarciaruiz, line 91: When we need remote /gnc/ekf we usually want the full rate. I'll likely remove this default rate from the configuration since Katie's on leave.

marinagmoreira: I can do that! Any topic that you usually want to define the rate from the default config?

rgarciaruiz: Thank you! Not right now. The rate will likely be useful for custom user topics. But that can be adjusted as needed.

marinagmoreira: ok done

PR #777, opened 2024-04-05
Convert executive to ros2

kbrowne15:
This executive code was converted to ros2 over a year ago. I'm not sure what state it is in. I thought it was compiling but not tested. Since it took quite a bit of work, I didn't want the work to get lost so whoever takes over converting the executive to ros2 should start here.


marinagmoreira:
closing as new PR includes these commits

PR #779, opened 2024-04-10
Remote docker vscode

marinagmoreira:
The PORT is still not configurable because to my knowledge you can't use variables in the devcontainer.json for forwardPorts. We can instead make a script that changes the port on all the files in the future


marinagmoreira:
@bcoltin @trey0 I tested this using beehive/bartlett. If one of you could test this to make sure it doesn't only work for me :)
 (ssh to beehive with vscode, open the astrobee folder in this branch, follow the instructions)

bcoltin (approved):

PR #780, opened 2024-04-11
Improve action result

marinagmoreira:
lab tested confirmed formatting and whether the error message is passed onto GDS


bcoltin (approved):
This is wonderful. Thanks for doing this.

PR #783, opened 2024-05-10
Release 0.17.3

marinagmoreira:
release procedure in https://babelfish.arc.nasa.gov/confluence/display/FFFSW/IRG-FFTEST302+-+2024-05-09+-+0.17.3+-+Astrobee+Release+Testing

- fixed pixel test that failed
- fixed obvious conversion issues with gnc_visualizer (doesn't influence the release procedure since it's not included in the debian). It opens up without errors, but plots not printing

### DIFF
--- a/tools/gnc_visualizer/scripts/visualizer.py
+++ b/tools/gnc_visualizer/scripts/visualizer.py
@@ -429,7 +429,7 @@ class Visualizer(QMainWindow):
                         self.menuBar().hide()
             elif event.type() == QtCore.QEvent.Leave and source is self:
                 self.menuBar().hide()
-        return super(Visualizer, self).eventFilter(source, event)
+        return super().eventFilter(source, event)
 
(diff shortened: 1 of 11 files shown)


tools/gnc_visualizer/scripts/visualizer.py
```diff
@@ -420,7 +429,7 @@ def eventFilter(self, source, event):
                         self.menuBar().hide()
             elif event.type() == QtCore.QEvent.Leave and source is self:
                 self.menuBar().hide()
-        return QtGui.QMainWindow.eventFilter(self, source, event)
+        return super(Visualizer, self).eventFilter(source, event)
```
trey0, line 432:
```suggestion
        return super().eventFilter(source, event)
```
Should be equivalent in Python 3.

bcoltin (approved):

PR #784, opened 2024-05-10
Update flashlight service name in scripts and docs

stephanie-eng:
As of #254, the flashlight service name was changed from:
`/hw/light_front/control` and `/hw/light_aft/control` (see [these lines from aacc34d](https://github.com/nasa/astrobee/blob/aacc34d89cf001d89ae68e2a9ebeefc13048ef2f/shared/ff_util/include/ff_util/ff_names.h#L434-L435))

to `/hw/flashlight_front/control` and `/hw/flashlight_aft/control` (see [these lines from develop](https://github.com/nasa/astrobee/blob/ce579297dcb91788f470e7d8ba373ed5b10f302f/shared/ff_common/include/ff_common/ff_names.h#L447-L448)).

Some scripts and docs reference the service name from before this change,


bcoltin (approved):
Thank you!

PR #785, opened 2024-05-10
Update brightness in setFlashlightBrightness example

stephanie-eng:
Using the executive to update the flashlight brightness requires the brightness be between [0 and 1](https://github.com/nasa/astrobee/blob/develop/management/executive/src/executive.cc#L2660-L2663).

Publishing the current example in the docs results in the command being rejected, as the brightness in the example of 100 is greater than 1.

This change updates the docs example to reflect this.


bcoltin (approved):
Thank you Stephanie!

PR #787, opened 2024-05-18
Fixed Gazebo IMU transform

rsoussan:
Currently the IMU sensor transform for gazebo is identity. This differs from the ISS bumble.config IMU transform, and this discontinuity causes the localizer to handle it incorrectly. Oddly, when changing the .xacro file though, gazebo doesn't seem to alter the IMU location. This was attempted even with crazy transforms that swap x with z axes, and this gave the same result in gazebo.


bcoltin:
Am I understanding right that this isn't working?

PR #797, opened 2024-06-25
Updated Map Matching With TEBLID and CLAHE

rsoussan:
Added TEBLID feature support for mapping and localization, added option for using CLAHE for histogram equalization, added options for adding similar images, best previous image, and essential matrix filtering.


localization/localization_node/src/localization.cc
```diff
@@ -27,55 +27,60 @@
...
+Localizer::Localizer(sparse_mapping::SparseMap* map): map_(map) {}
+
+void Localizer::ReadParams(config_reader::ConfigReader& config) {
+  camera::CameraParameters cam_params(&config, "nav_cam");
+  std::string prefix;
+  const auto detector_name = map_->GetDetectorName();
+  if (detector_name == "ORGBRISK") {
+    prefix = "brisk_";
+  } else if (detector_name == "TEBLID") {
+    prefix = "teblid_";
```
trey0, line 39: (Possibly support "TEBLID512" and "TEBLID256"?)

localization/localization_node/src/localization.cc
```diff
@@ -27,55 +27,60 @@
...
+  LOAD_PARAM(min_features, config, prefix);
+  LOAD_PARAM(max_features, config, prefix);
 
   // This check must happen before the histogram_equalization flag is set into the map
   // to compare with what is there already.
   sparse_mapping::HistogramEqualizationCheck(map_->GetHistogramEqualization(),
-                                             histogram_equalization);
+                                             loc_params.histogram_equalization);
+  // Check consistency between clahe params
+  if (loc_params.use_clahe && (loc_params.histogram_equalization != 3 || map_->GetHistogramEqualization() != 3)) {
```
trey0, line 76: The literal 3 here is a bit obscure. Probably there should be an enum or at least a macro that you can compare to?

localization/sparse_mapping/include/sparse_mapping/sparse_map.h
```diff
@@ -272,15 +229,12 @@ struct SparseMap {
   std::vector<cv::Mat> cid_to_descriptor_map_;
   // generated on load
   std::vector<std::map<int, int> > cid_fid_to_pid_;
+  std::vector<std::map<int, int>> cid_to_matching_cid_counts_;
```
trey0, line 232: If we're adding new fields to the map, ideally we should also log their memory usage in our memory debug statistics. (Not sure if you were aware there's per-field code for that.)

trey0 (commented):
Hi, Ryan- Not sure I should presume to approve because I'm pretty out of the loop on some of the integration issues. This is looking great so far! Just a few questions / minor comments.

localization/interest_point/include/interest_point/BAD.h
```diff
@@ -0,0 +1,99 @@
+/**
+ * @copyright 2021 Xoan Iago Suarez Canosa. All rights reserved.
+ * Constact: iago.suarez@thegraffter.com
+ * Software developed in the PhD: Low-level vision for resource-limited devices
```
bcoltin, line 4: what license?

bcoltin: and does this really need to be included instead of linked to

localization/localization_node/src/localization.cc
rsoussan: Done

bcoltin (approved):
Fix headers like we discussed and any minor things and if you've tested looks good.

localization/interest_point/include/interest_point/BAD.h
rsoussan: Done

astrobee/config/localization.config
```diff
@@ -15,17 +15,51 @@
...
-min_brisk_threshold = 20.0
-default_brisk_threshold = 90.0
-max_brisk_threshold = 110.0
 matched_features_on = false
 all_features_on = false
+num_threads = 1
+verbose_localization = false
+visualize_localization_matches = false
+
+-- BRISK
```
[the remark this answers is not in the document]
rsoussan: Added teblid512 and teblid256, we don't really support SURF for loc since we never merged the vocab branch so I'll leave that out

localization/interest_point/src/BAD.cpp
```diff
@@ -0,0 +1,611 @@
+/**
+ * @copyright 2021 Xoan Iago Suarez Canosa. All rights reserved.
+ * Constact: iago.suarez@thegraffter.com
+ * Software developed in the PhD: Low-level vision for resource-limited devices
+ */
```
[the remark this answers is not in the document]
rsoussan: (Used similar link with license as suggested by @bcoltin)

localization/localization_node/src/localization.cc
rsoussan: Done

localization/sparse_mapping/include/sparse_mapping/sparse_map.h
rsoussan: Oh actually I wasn't aware of that, which file(s) should I edit?

rsoussan: (found it, looks like it's still in PR, I'll update the pr)

PR #801, opened 2024-07-03
Support radial fisheye colmap distortion.

bcoltin:
Specify just two parameters in the config file to use this model.

### DIFF
--- a/localization/camera/src/camera_params.cc
+++ b/localization/camera/src/camera_params.cc
@@ -229,6 +229,8 @@ void camera::CameraParameters::SetDistortion(Eigen::VectorXd const& distortion) 
     // Inside tangent function
     distortion_precalc2_ = 2 * tan(distortion[0] / 2);
     break;
+  case 2:
+    // Fall through intended.
   case 4:
     // Fall through intended.
   case 5:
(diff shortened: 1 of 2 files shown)


localization/camera/src/camera_params.cc
```diff
@@ -266,6 +268,27 @@ void camera::CameraParameters::DistortCentered(Eigen::Vector2d const& undistorte
...
+    double k1 = distortion_coeffs_[0];
+    double k2 = distortion_coeffs_[1];
+
+    // To relative coordinates
+    Eigen::Vector2d norm = undistorted_c.cwiseQuotient(focal_length_);
+    double r = norm.norm();
+    if (r > std::numeric_limits<double>::epsilon()) {
+      double theta = atan(r);
+      double theta2 = theta * theta;
+      double thetad = theta * (1 + k1 * theta2 + k2 * theta2 * theta);
```
rsoussan, line 282: I think this should read ".....k2*theta2*theta2"

bcoltin: Good catch, thanks! Fixed

rsoussan (changes requested):
Looks good aside from the typo

rsoussan (approved):

PR #804, opened 2024-07-30
Sparse map multiple cameras

rsoussan:
Add support for using multiple cameras when making sparse maps. Updated localization node to localize with the camera params from the current robot config. 
@amoravargas


bcoltin (approved):
Assuming you tested with both new and previous maps. Impressed you managed to finish this in time, great work!

rsoussan:
> Assuming you tested with both new and previous maps. Impressed you managed to finish this in time, great work!

Thanks! Yeah I tested map creation with different camera models, including merging, and localizing on maps old and new. Looks good, we might want to have @amoravargas test it with a real mapping pipeline before we merge, also haven't done granite tests

bcoltin:
Make sure to fix the CI too

PR #807, opened 2024-07-31
Colmap Instructions and Scripts

bcoltin:
Readme with instructions and shortcomings, scripts for trying to do fancier things that I wasn't able to get fully working in time.


localization/sparse_mapping/scripts/colmap/incremental.py
```diff
@@ -0,0 +1,314 @@
...
+            image_list,
+            input_path=output_path,
+        ) as cfg:
+            os.system(
+                "colmap mapper --project_path %s > %s/colmap.out 2>&1"
+                % (cfg.file_name(), output_path)
+            )
+
+    def create_leaf_models(self, images, subdir="leaf"):
+        if type(images) is dict:
```
trey0, line 74: `if isinstance(images, dict):`

trey0 (commented):
Hi, Brian- I'm not familiar enough with colmap to truly understand what's going on here so won't presume to approve, but I had a bunch of minor comments about Python style. This looks like it was a lot of work!

trey0:
Ah, one more general comment is that `pylint` is really helpful. I find some of its warnings silly and don't bother to fix everything it reports, but it's usually worth running anyway. I'm pretty sure it would have caught several of the things I commented on.

kiruthikpurpose (approved):
Not gonna create a big difference. But the order of the packages becomes less confused for humans to check out what they could be missing when cross-compared.

kiruthikpurpose (changes requested):
Change the file name from readme.md to README.md

Look for neat organized structure for the documentation in README.md

PR #812, opened 2024-10-17
Correct OF nodelet type name so the system monitor can load it

rgarciaruiz:


rgarciaruiz:
One of the tests fail because #811 fixes that. Should we merge anyway or wait until #811 is merged to develop?

kbrowne15 (approved):

kbrowne15:
It's up to you. I did approve #811 so if you want to merge that in first, you can. Otherwise, the failed test doesn't relate to this pr so I'm okay with you merging it in even with a failing test.

PR #825, opened 2025-07-15
Updated Scripts For the Ground Truth Process

Cmaso010:
The auto_run.py allows the user to run the ground truth script on a list of bags. The bag_to_csv.py takes the bag generated from ground truth script and extracts position, orientation, and time and formats them into a csv file. The make_map.py has been updated to look in all necessary directories for images that the map is reliant on instead of assuming all images are in the same directory. The dense_model.py creates dense models from colmap for the validation process


tools/localization_analysis/scripts/auto_run.py
```diff
@@ -0,0 +1,44 @@
...
+parser.add_argument(
+    "--bag_path", "-bp", type=str, help="path of the directory containing your bags"
+)
+
+parser.add_argument(
+    "--map_path", "-m", type=str, help="path of the directory containing your map"
+)
+
+parser.add_argument(
+    "--bags", "-b", nargs="+", type=str, help="List of bag names to process"
```
bcoltin, line 22: I would make bag_path -b, and turn --bags into a positional argument (meaning you don't need a flag at all, just put the files on the command line standalone)

bcoltin (changes requested):

bcoltin (approved):

